PR #89, opened 2020-06-28
Switch to Buidler for running CI tests

alsco77:
- [x] Add basic Buidler
- [x] Fix broken tests (error messages different - specify the evm release?)
- [x] Hook in gas reporter (https://buidler.dev/plugins/buidler-gas-reporter.html)
- [x] Use Buidler for coverage too (https://buidler.dev/plugins/solidity-coverage.html)
- [x] Figure out what to do with truffle
- [x] Optimise the coverage time (or revert to truffle)


cgewecke:
Hi @alsco77 

Just a quick note...there's an experimental version of buidler-gas-reporter that uses BuidlerEVM directly which might work for you. It's faster and may be easier to get running.

It only works for truffle contract based projects (but you're in that category)

```
npm install --save-dev buidler-gas-reporter@beta
```

alsco77:
Hey @cgewecke, thanks for that, i'll try that out.

Since you are here - you can see that the Coverage build is failing with a timeout. Not sure why this would be failing on the `buidler coverage` and not `truffle coverage`.. since they both seem to be relying heavily on the core `solidity-coverage` package.

cgewecke:
@alsco77 Have to run to a meeting super quick but will circle back here in 1 hr. and look...

alsco77:
Tweaked a few things - there were a ALOT of calls happening in one specific test.

Just re-adding the max cache size flag for the coverage to speed it up, as it seems to be less performant vs historic builds with truffle.

cgewecke:
> tried adding the beta version but it was just giving me the same error when trying to use the buidlerevm

Oh yes, makes sense sorry. I should be able to fix that super quick...

cgewecke:
@alsco77 Sorry - actually I confused solidity-coverage and the gas-reporter just now. 

What error are you seeing with the reporter@beta?

alsco77:
Thanks for your attention @cgewecke, it's greatly appreciated. You were right on removing the `eth-gas-reporter`. I tried that and ran a few `buidler test ./test/TestFile.spec.ts` and it works - but when running the full suite, it seems to produce a large amount of errors. See the most recent commit on this PR (causing the tests to go from passing to failing) and the CircleCI build output below which shows a lot of event listener errors: 

image

cgewecke:
@alsco77 Ah thanks for trying it! Will see if I can fix it sometime today.

alsco77:
CI run time reduced by ~50%! :fire: 

image

PR #91, opened 2020-06-30
Upgrade buidler-gas-reporter to 0.1.4-beta.4

cgewecke:
@alsco77 Oh sorry..wanted to double-check this in CI here but it looks like it's not enabled for forks. 

Please feel free to close this.

#89 Ran clean for me locally with this update. 

The reporter may still slow things down a little - there's a boolean `enabled` option for the buidler config you can set conditionally with an env variable if you'd rather not run this all the time.


alsco77:
Thanks for that. We can see how the CI runs on the main branch and then i'll look into enabling CI for forks in the future. Left it off initially to avoid potentially spamming our limits :+1:

PR #98, opened 2020-08-09
Implemented period based yield analysis on SavingsManager

alsco77:
**Upgrade type**: MODULE REPLACEMENT VIA NEXUS
**Affected file**: `SavingsManager.sol`
**Version**: 1.1

### Description

Address bug originally submitted through Bug Bounty - See [Devan - Bug bounty response.pdf](https://github.com/mstable/mStable-contracts/files/5048099/Devan.-.Bug.bounty.response.pdf)

Removes the blocker on collecting interest more than once in 30 minute period.

Retains existing 'extrapolatedAPY' calculations when it has been longer than 30 minutes since the last interest collection.

If it has been less than 30 minutes, it simply checks that the supply has not inflated by more than 0.1% (or 1e15) during that PERIOD (30 mins). At a 0.1% 'SWAP' fee, this would mean that the total supply of the mAsset would need to be swapped between the two collections, if this were to be hit.

**Mechanism**:

```
        //    If it has been 30 mins since last collection, reset period data
        //    Calculate APY from lastCollection, interest, newTotalSupply (like before)
        if(timeSinceLastCollection > THIRTY_MINUTES) {
            lastPeriodStart[_mAsset] = now;
            periodYield[_mAsset] = 0;
        }
        //    Else if period has elapsed, start a new period from the lastCollection time
        //    Its under 30 mins since last collection, so we can calculate the increase
        //    and protect it from going over 10 bps
        else if(timeSincePeriodStart > THIRTY_MINUTES) {
            lastPeriodStart[_mAsset] = previousCollection;
            periodYield[_mAsset] = interestCollected;
        }
        //    Else add yield to period yield
        //    We use the full yield for the period in the 10 bps calculation
        else {
            inflationOperand = periodYield[_mAsset].add(interestCollected);
            periodYield[_mAsset] = inflationOperand;
        }
```

```
        //   0        30        60        90        120       
        //   | - - - - | - - - - | - - - - | - - - - |
        //   ^            ^   ^    ^    ^            ^
        //  start        40   50  65    80          120
        //  @time - Description (periodStart, lastCollection, periodYield)
        //  @40 - Should start new period (40, 40, 0)
        //  @50 - Should calc in period 40-70 (40, 50, X)
        //  @65 - Should calc in period 40-70 (40, 65, X+Y)
        //  @80 - Should start new period from last collection, 65-95 (65, 80, Z)
        //  @120 - Should start new period (120, 120, 0)
```

### Deployment plan

- [x] Write and comprehensively test the upgrade
- [x] Get external PR review from specialist
  - [x] 1 approval
  - [x] 2 approval
- [x] Deploy `SavingsManager v1.1` to Ropsten/Mainnet
- [x] Use [DiffChecker](https://www.diffchecker.com/) to validate the prev and future contracts once verified
- [x] Call `proposeUpgrade` on the `Nexus` with implementation addresses
- [x] Verify proposal is in the Nexus and the keys match up
- [x] Accept on Ropsten and test
- [x] Wait 1 week
- [x] Accept on Mainnet FORK by unlocking the gov address
- [x] Accept on Mainnet

#### Mainnet Details

Deployed address: `0x7046b0bfc4c5eeb90559c0805dd9c1a6f4815370`

#### Ropsten Details

Deployed address: `0x755629d1ecad92db2b81a3a88207bf6d2b864136`

### DIFF
--- a/contracts/savings/SavingsManager.sol
+++ b/contracts/savings/SavingsManager.sol
@@ -19,8 +19,8 @@ import { StableMath } from "../shared/StableMath.sol";
  * @author  Stability Labs Pty. Ltd.
  * @notice  Savings Manager collects interest from mAssets and sends them to the
  *          corresponding Savings Contract, performing some validation in the process.
- * @dev     VERSION: 1.0
- *          DATE:    2020-03-28
+ * @dev     VERSION: 1.1
+ *          DATE:    2020-07-29
  */
 contract SavingsManager is ISavingsManager, PausableModule {
 
@@ -36,12 +36,13 @@ contract SavingsManager is ISavingsManager, PausableModule {
     event InterestCollected(address indexed mAsset, uint256 interest, uint256 newTotalSupply, uint256 apy);
     event InterestDistributed(address indexed mAsset, uint256 amountSent);
     event InterestWithdrawnByGovernor(address indexed mAsset, address recipient, uint256 amount);
-    event WaitFor(uint256 timeRemaining);
 
     // Locations of each mAsset savings contract
     mapping(address => ISavingsContract) public savingsContracts;
     // Time at which last collection was made
+    mapping(address => uint256) public lastPeriodStart;
     mapping(address => uint256) public lastCollection;
+    mapping(address => uint256) public periodYield;
 
     // Amount of collected interest that will be sent to Savings Contract (100%)
     uint256 private savingsRate = 1e18;
@@ -49,6 +50,7 @@ contract SavingsManager is ISavingsManager, PausableModule {
     uint256 constant private SECONDS_IN_YEAR = 365 days;
     // Theoretical cap on APY to avoid excess inflation
     uint256 constant private MAX_APY = 15e18;
+    uint256 constant private TEN_BPS = 1e15;
     uint256 constant private THIRTY_MINUTES = 30 minutes;
 
     constructor(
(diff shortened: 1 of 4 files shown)


dpurhar27:
Overall, this looks good besides the one comment I made above. Removing the time barrier before collecting interest and always updating the exchange rate fixes the original issue 👍.

Based on the provided example:

> this would mean that the total supply of the mAsset would need to be swapped between the two collections

I think the chosen limit of 0.1% isn't too restrictive as long as the AUM is high relative to trade volume.

_Note, copying my original comment over from the private repo_

contracts/savings/SavingsManager.sol
```diff
@@ -136,58 +138,96 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+        // e.g. 0.01% (1e14 * 1e18) / 2.74..e15 = 3.65e16 or 3.65% apr
+        // e.g. (4.1667e12 * 1e18) / 5.7..e13 = 7.1e16 or 7.1% apr
+        // e.g. (1e16 * 1e18) / 50e18 = 2e14
+        extrapolatedAPY = percentageIncrease.divPrecisely(yearsSinceLastCollection);
+
+        //      If over 30 mins, extrapolate APY
+        if(_timeSinceLastCollection > THIRTY_MINUTES) {
+            require(extrapolatedAPY < MAX_APY, "Interest protected from inflating past maxAPY");
+        } else {
+            require(percentageIncrease < TEN_BPS, "Interest protected from inflating past 10 Bps");
```
dpurhar27, line 230: It seems the goal is to make sure the supply isn't inflated by 0.1% or more if the time between collections is 30 minutes or less.

However, the current implementation only checks this atomically per collection. It doesn't consider the past inflation already experienced in the last _X_ minutes.

**Example:** Bob decides he wants to corrupt mStable's asset through inflation. He exerts massive trading volume. To bypass the `MAX_APY` check, he makes sure he's updating the `timeSinceLastCollection` within 30 minutes between each collection. 
Now, he makes sure he's only inflating the supply by 0.09%, then updating the `timeSinceLastCollection` (through calling the `collectAndDistributeInterest()`). He can now inflate the asset by over 2.7% in 30 minutes (assuming he does this every minute), and bypass any restrictive mechanisms on inflation that are set.

That example uses a malicious actor, but there's natural ways this could happen too. The point being, if I'm understanding the reasoning of the restriction, the implementation should consider the past inflation too when it's been 30 minutes or less since last collection.

A state variable could be added that get's incremented by the inflation experienced as long as it's in one 30 minute collection time. Once the collection time overflows 30 minutes, it can be zeroed out and restarted. This state variable is what would be used in the above check.

`require(periodPercentageIncrease < TEN_BPS, "...");`

Where the `period...` variable is referring to the sum of inflation in the current 30 minute collection period.

_Note, copying my original comment over from the private dev. repo_

dpurhar27:
Overall, this looks good besides the one comment I made above. Removing the time barrier before collecting interest and always updating the exchange rate fixes the original issue 👍.

Based on the provided example:

> this would mean that the total supply of the mAsset would need to be swapped between the two collections

I think the chosen limit of 0.1% isn't too restrictive as long as the AUM is high relative to trade volume.

_Note, copying my original comment over from the private dev. repo_

contracts/savings/SavingsManager.sol
alsco77: This comment is outdated due to the summing of the period yield in line 168

PR #101, opened 2020-08-23
Added prerequisites and set block gas limit

shapeshed:
I ran into issues using Node.js v13.10.1 and found that that the LTS v10.22.0 release works. I also found that the default block gas limit is too low for the setting in `truffle-config.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ Found a bug? Claim a reward from our open [Bug Bounty](https://docs.mstable.org/
 
 ## Dev notes
 
+### Prerequisites
+
+* Node.js v10.22.0 (you may wish to use [nvm][1])
+* [ganache-cli][2]
+
 ### Installing dependencies
 
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,7 +64,7 @@ Deployment scripts are located in `migrations/src`. To run, start `ganache` or `
 *NB: You should locally use the latest version of `ganache-cli`, as contracts rely on recent opcodes*
 
 ```
-$ ganache-cli -p 7545
+$ ganache-cli -p 7545 -l 2000000
```
alsco77, line 67: Should be `ganache-cli -p 7545 -l 8000000`, 2 million gas is too low and wouldn't allow the contracts to be deployed

alsco77 (changes requested):
Thanks for the PR but requires a tweak to go through

README.md
shapeshed: Thanks, have updated it.

alsco77 (approved):

PR #110, opened 2020-09-21
Liquidator

shapeshed:


contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    )
+
+        external
+        onlyGovernor
+    {
+        require(liquidations[_bAsset].bAsset != address(0), "No liquidation for this bAsset");
+
+        Liquidation storage liq = liquidations[_bAsset];
+        liq = liquidations[_bAsset];
+        liq.bAsset = _bAsset;
```
alsco77, line 157: redundant

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        liq.integration = _integration;
```
alsco77, line 158: We should probably only change this if there are no outgoing tokens still left to be collected

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    * @dev Delete a liquidation
+    * @param _bAsset The _bAsset for the liquidation
+    */
+    function deleteLiquidation(address _bAsset)
+        external
+        onlyGovernor
+    {
+        require(liquidations[_bAsset].bAsset != address(0), "No liquidation for this bAsset");
+
+        delete liquidations[_bAsset];
```
alsco77, line 177: We should only delete if the liquidator does not own any sellToken, and if there are no more rewards to be collected.

There is a risk of having both sellTokens and other locked in here

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    /**
+    * @dev Allows a calling integration contract to collect redeemed tokens
+    */
+    function collect(address _bAsset)
+        external   
+    {
+        Liquidation memory liq = liquidations[_bAsset];
+        require(msg.sender == liq.integration, "Only integration contract can execute");
+
+        address sendAddress = liq.integration;
```
alsco77, line 205: We can just use `msg.sender` instead of reading the `integration` here to save gas. OR read `liq.integration` once and save it to a local var, instead of reading twice.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+
+    /**
+    * @dev Allows a calling integration contract to collect redeemed tokens
+    */
+    function collect(address _bAsset)
+        external   
+    {
+        Liquidation memory liq = liquidations[_bAsset];
+        require(msg.sender == liq.integration, "Only integration contract can execute");
+
```
alsco77, line 204: Missing: timer to avoid frequent collections. We should not collect EVERY time this is called - but once per hour (with some added randomness)

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        address sendAddress = liq.integration;
+        address asset = liq.sellToken;
+        
+        uint256 balance = IERC20(asset).balanceOf(address(this));
```
alsco77, line 208: We don't want to send all the balance at once.. we wanted to send ~20% or so, or say, minimum of $200 worth. This would just send it all

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        IERC20(asset).safeApprove(sendAddress, 0);
+        IERC20(asset).safeApprove(sendAddress, balance);
```
alsco77, line 210: Don't need to approve

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        IERC20(asset).safeTransfer(sendAddress, balance);
```
alsco77, line 211: We are wanting to send the `pToken`, not the `bAsset`. The bAsset should be deposited in the corresponding lending market during the `triggerLiquidation` function and stored here for dripping to collection.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    /**
+    * @dev Triggers a liquidation
+    * @param _bAsset The _bAsset liquidation to be triggered
+    */
+    function triggerLiquidation(address _bAsset)
+        external
+    {
+        Liquidation memory liquidation = liquidations[_bAsset];
+
+        require(liquidation.lastCalled + 1 days > now, "Trigger liquidation only callable every 24 hours");
```
alsco77, line 223: This logic is the wrong way round. Also, use `block.timestamp` over `now`

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+
+        address asset = liquidation.sellToken;
+        uint256 balance = IERC20(asset).balanceOf(address(this));
+        require(balance != 0, "No tokens to liquidate");
+
+        uint256 sellAmount;
+
+        uint256 randomAmountToSell = uint(blockhash(block.number-1)) % 3000 + 1000;
```
alsco77, line 231: This is going to sell between 1000 & 4000 of `asset`. This is 4000 COMP. This is far too much (maybe it would make sense in USD terms)

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+
+        if (balance > sellAmount) {
```
alsco77, line 233: sellAmount is always 0, so this will always be true. You should turn `randomAmountToSell` (USD) into units as per some uniswap oracle, then set this to sellAmount (units).

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+
+        IUniswapV2Router02(uniswapAddress).swapExactTokensForTokens(
+            sellAmount,     
+            uint256(0),
+            liquidation.uniswapPath,
+            address(this),
+            now.add(1800)
+        );
+
+        liquidation.lastCalled = block.timestamp;
```
alsco77, line 250: This should be moved up to the top of the function. Better to do at top of func to avoid reentrancy attacks

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        IERC20(asset).safeApprove(uniswapAddress, sellAmount);
+
+        IUniswapV2Router02(uniswapAddress).swapExactTokensForTokens(
+            sellAmount,     
+            uint256(0),
+            liquidation.uniswapPath,
+            address(this),
+            now.add(1800)
+        );
+
```
alsco77, line 249: We need to deposit the newly purchased bAsset into the corresponding lending market here, so that the pToken is ready for collection

alsco77 (changes requested):

contracts/masset/liquidator/Liquidator.sol
shapeshed: yeah this is a point I wanted to discuss about how best to determine the spot price and compute this.

contracts/masset/liquidator/Liquidator.sol
alsco77: can do something similar to `cToken to bAsset` calculation

contracts/interfaces/ILiquidator.sol
```diff
@@ -0,0 +1,34 @@
+pragma solidity 0.5.16;
+pragma experimental ABIEncoderV2;
+
+/**
+  * @title ILiquidator
+  * @dev Interface for the Liquidator component
+  */
+
+interface ILiquidator {
+    struct Liquidation {
```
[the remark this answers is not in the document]
shapeshed: Removed the interface for now. Will move this into an interface as part of refactoring when tests are passing.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    event UniswapUpdated(address indexed uniswapAddress);
+
+    address public uniswapAddress;
+
+    struct Liquidation {
+        address     bAsset;
+        address     integration;
+        address     sellToken;
+        address[]   uniswapPath;
+        uint        amount;
```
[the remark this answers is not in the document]
shapeshed: is the USD (specifically the USD stablecoin amount)

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+        address     _sellToken,
+        address[]   calldata _uniswapPath,
+        uint        _amount,
+        bool        _paused
+    )
+        external
+        onlyGovernor
+    {
+        require(_bAsset != address(0), "_bAsset cannot be zero address");
+        require(_integration != address(0), "integration cannot be zero address");
```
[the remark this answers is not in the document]
shapeshed: Added a check for this. Commented for now until mocks are added to tests.

contracts/masset/liquidator/Liquidator.sol
shapeshed: Added a check for this, commented until I add a mock for tests.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,263 @@
...
+    * @param _bAsset The _bAsset address that this liquidation is for
+    * @param _integration The integration contract address for the _bAsset
+    * @param _sellToken The token address to be sold
+    * @param _uniswapPath The Uniswap path as an array of addresses e.g. [COMP, WETH, DAI]
+    * @param _amount The amount of tokens to be sold
+    * @param _paused Whether the liquidation is paused
+    */
+    function createLiquidation(
+        address     _bAsset,
+        address     _integration,
```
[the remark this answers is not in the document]
shapeshed: Will add an enum on the struct for this

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -16,6 +16,8 @@ contract CompoundIntegration is InitializableAbstractIntegration {
     event RewardTokenCollected(address recipient, uint256 amount);
     event SkippedWithdrawal(address bAsset, uint256 amount);
 
+    uint256 internal lastClaimed;
```
shapeshed, line 19: Does this work with the upgradeable pattern?

alsco77: Yes, it will be initialised with the value 0, so should execute on the first deposit/withdrawal

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+     * @dev Checks whether a claim should be made
+     * This compares the block.timestamp with a somewhat random time
+     * Adds randomness by muliplying the 1 hour delay between 1x and 3x
+     * @param _bAsset Address for the bAsset
+     */
+    function _checkClaim(address _bAsset)
+        internal
+    {
+        uint256 salt = uint256(keccak256(abi.encodePacked(blockhash(block.number)))).mod(3000000);
+        uint256 timeDelay = ((uint256(1 hours)).mul(salt)).div(1000000);
```
alsco77, line 261: What number is this supposed to produce?

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+     */
+    function _claim(address _bAsset)
+        internal
+    {
+        lastClaimed = block.timestamp;
+
+        address liquidator = nexus.getModule(keccak256('Liquidator'));
+        address cToken = bAssetToPToken[_bAsset];
+        require(cToken != address(0), "cToken does not exist");
+        uint256 allowance = IERC20(cToken).allowance(liquidator, address(this));
```
alsco77, line 282: Are we assuming the allowance is always == to the total amount of cTokens held by the liquidator?

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+
+        if (allowance > 0 && allowance < 20) {
```
alsco77, line 284: This `allowance` is in cToken base units. I'm assuming this is trying to check if there is between 0 & 20 bAssets to give? (or rather, do the transfer if less than $1k resides there). If we want to check for a dollar amount, we need to: a) convert the cToken allowance into bAsset units, b) determine how many decimals the bAsset has so we can figure out the $$ value of the allowance

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+            IERC20(cToken).safeTransferFrom(liquidator, address(this), allowance);
+        }
+
+        if (allowance > 0) {
```
alsco77, line 288: Should be else if, otherwise it will be called twice

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+            uint256 randomBp = uint256(blockhash(block.number-1)).mod(3000).add(1000);
+            uint256 toTransfer = allowance.mul(randomBp).div(uint(10000));
```
alsco77, line 290: To confirm - this sends between 10 & 40% of the cTokens

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+        address     bAsset;
+        address     integration;
+        address     sellToken;
+        address[]   uniswapPath;
+        uint        amount;
+        bool        paused;
+        uint        lastTriggered;
+    }
+
+    mapping(address => Liquidation) public liquidations;
```
alsco77, line 59: Given that there is only 1 liquidation per `Integration`, the mapping should be `integration` centric and not `bAsset` centric. Should probably change this. After which, there is no need to store the actual integration address in the struct - that is double saving.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+        bool        _paused
+    )
+        external
+        onlyGovernor
+    {
+        require(_bAsset != address(0), "_bAsset cannot be zero address");
+        require(_integration != address(0), "integration cannot be zero address");
+        // TODO add mocks for this in tests
+        //address pToken = IPlatformIntegration(_integration).bAssetToPToken(_bAsset);
+        //require(pToken != address(0), "no pToken for this bAsset");
```
alsco77, line 112: Also store the pToken

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+
+        emit LiquidationCreated(_bAsset);
+    }
+
+    /**
+     * @dev Get a liquidation
+     * @param   _bAsset Address for the underlying bAsset
+     * @return  liquidation The liquidation data
+     */
+    function readLiquidation(address _bAsset)
```
alsco77, line 130: No real need for this given that the info is on a public var and we can read off chain anyway

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +245,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+       amount = _underlying.mul(1e18).div(exchangeRate);
+    }
+
+    /**
+     * @dev Checks whether a claim should be made
+     * This compares the block.timestamp with a somewhat random time
+     * Adds randomness by muliplying the 1 hour delay between 1x and 3x
+     * @param _bAsset Address for the bAsset
+     */
+    function _checkClaim(address _bAsset)
```
alsco77, line 257: We are only selling 1 asset on the Liquidator, so checking the claim for other bassets here will result in nothing being collected. Either we need to decide to sell more than 1 asset on the Liquidator, or specifically only collect for 1 bAsset (dai) here.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+                "Trigger liquidation only callable every 24 hours");
+
+        // Token being sold is the first in the Uniswap path
+        address sellToken = liquidation.uniswapPath[0];
+
+        // Token being bought is the last in the Uniswap path
+        address buyToken = liquidation.uniswapPath[liquidation.uniswapPath.length.sub(1)];
+
+        // Get allowance given by the integration contract for the sell token
+        uint256 allowance = IERC20(sellToken).allowance(liquidation.integration, address(this));
```
alsco77, line 241: If we give infinite approval from the CompoundIntegration, then this allowance will always be a high number. It is the balance we are supposed to be checking for - does the Integration have any new COMP to distribute

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+        if (allowance > 0) {
+            IERC20(sellToken).safeTransferFrom(liquidation.integration, address(this), allowance);
+        }
+
+        uint256 balance = IERC20(sellToken).balanceOf(address(this));
+        require((balance > 0), "No sell tokens to liquidate");
+
+        // The amount we want to receive
+        // This computes a randomised amount of the buyToken wanted between 1000 & 4000
+        uint256 randomAmountWanted = uint256(blockhash(block.number-1)).mod(3000).add(1000);
```
alsco77, line 253: This will return a number between 1000-4000, which in bAsset units is barely anything.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,303 @@
...
+
+        // Get minimum amount of sellTokens needed for the amount of buyTokens
+        // https://uniswap.org/docs/v2/smart-contracts/router02/#getamountsout
+        uint[] memory amountsIn = IUniswapV2Router02(uniswapAddress).getAmountsIn(randomAmountWanted, liquidation.uniswapPath); 
+
+        // Add 10 % to remain over minimum amount
+        uint256 tenPercentOfMinAmount = amountsIn[0].mul(uint(1000)).div(uint(10000));
+        uint256 minAmount = amountsIn[0].add(tenPercentOfMinAmount);
```
alsco77, line 261: What is the purpose of adding 10% to the amount here?

alsco77 (changes requested):

contracts/masset/liquidator/Liquidator.sol
shapeshed: The `Integration` interface doesn't have a way to read the `bAsset` and a lot of the logic is around the underlying `bAsset` so I favour leaving it this way. Plus if the integration contract changes for a bAsset this is supported (e.g. using Aave instead of Compound for DAI).

contracts/masset/liquidator/Liquidator.sol
shapeshed: Was mainly using it for testing. Can you recommend a way to read values that are set without setting an external function for tests?

contracts/masset/liquidator/Liquidator.sol
shapeshed: Added checks for unclaimed pTokens and unsold sellTokens

contracts/masset/platform-integrations/CompoundIntegration.sol
shapeshed: 1 hour with a multiplier between 1 and 3

contracts/masset/platform-integrations/CompoundIntegration.sol
shapeshed: yes the contract assumes infinite approval

contracts/masset/platform-integrations/CompoundIntegration.sol
shapeshed: Yes added a comment

alsco77 (commented):
There are a lot of linting errors. This is something that should be being picked up by editor as going along.

Other review given off chain

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,329 @@
...
+        address         _sellToken,
+        uint            _trancheAmount,
+        LendingPlatform _lendingPlatform,
+        address[]       calldata _uniswapPath,
+        bool            _paused
+    )
+        external
+        onlyGovernor
+    {
+        require(_bAsset != address(0), "bAsset cannot be zero address");
```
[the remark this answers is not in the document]
shapeshed: Added a check that `integration` is not set on the struct as this is required on create.

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,329 @@
...
+        require(_integration != address(0), "integration cannot be zero address");
+        require(_sellToken != address(0), "sellToken cannot be zero address");
+        require(_trancheAmount != uint(0), "trancheAmount cannot be zero");
+        require(_uniswapPath.length >= uint(2), "uniswapPath must have at least two addresses");
+
+        address pToken = IPlatformIntegration(_integration).bAssetToPToken(_bAsset);
+        require(pToken != address(0), "no pToken for this bAsset");
```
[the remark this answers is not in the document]
shapeshed: Added approval to `createLiquidation` 111-112 and `updateLiquidation` 159-160

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,329 @@
...
+    * @dev Deposits to Compound 
+    * @param _pToken The _pToken to mint
+    * @param _bAsset The _bAsset liquidation to be triggered
+    */
+    function depositToCompound(address _pToken, address _bAsset)
+        internal
+    {
+        uint256 bAssetBalance = IERC20(_bAsset).balanceOf(address(this));
+        require((bAssetBalance > 0), "No tokens to deposit");
+        require(ICERC20(_pToken).mint(bAssetBalance) == 0, "cToken mint failed");
```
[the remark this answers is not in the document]
shapeshed: Approval given in `createLiquidation`

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -0,0 +1,329 @@
...
+        internal view returns (uint256, uint256)
+    {
+
+        // The _trancheAmount is the number of bAsset we want to buy each time
+        // DAI has 18 decimals so 1000 DAI is 10*10^18 or 1000000000000000000000
+        // This value is set when adding the liquidation
+        // We randomize this amount by buying betwen 80% and 95% of the amount.
+        // Uniswap gives us the amount we need to sell with `getAmountsIn`.
+        uint256 randomBasisPoint = uint256(blockhash(block.number-1)).mod(uint(1500)).add(uint(8000));
+        uint256 amountWanted = _trancheAmount.mul(randomBasisPoint).div(uint(10000));
```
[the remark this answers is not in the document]
shapeshed: This is intended to randomise the tranche amount. So if `_trancheAmount` is 1000 DAI it will generate a number between 800 & 950 to send to Uniswap to find out how much COMP to sell. If we don't care about the same `_trancheAmount` each time these lines are obsolete.

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -239,6 +246,55 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+       amount = _underlying.mul(1e18).div(exchangeRate);
+    }
+
+    /**
+     * @dev Checks whether a claim should be made
+     * This compares the block.timestamp with a somewhat random time
+     * Adds randomness by muliplying the 1 hour delay between 1x and 3x
+     * @param _bAsset Address for the bAsset
+     */
+    function checkClaim(address _bAsset)
```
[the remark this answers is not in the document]
shapeshed: We can support an integration having more than one liquidation as `_bAsset` is the key on the Liquidations mapping on the Liquidator. This works I think. Liquidations share the same balance of COMP but will sell for different Uniswap paths `COMP > WETH > DAI` or `COMP > WETH > USDC`.  In this case there are two `TriggerLiquidation` calls to the liquidator with the `bAsset` as the key that will sell COMP per the `trancheAmount` and the `uniswapPath` for the liquidiation. On the `CompoundIntegration` side the `bAsset` is used on deposits and withdrawals to lookup the `pToken` and transfer any tokens back to itself, so it works for supporting two liquidations on both sides unless I am missing your point. 

On the `CompoundIntegration` side we would need to change `lastClaimed` to be a mapping to take account of multiple `bAssets`. 

```
mapping(address => uint256) public lastClaimed;
```

If you don't see any problems with supporting multiple `bAssets` I suggest we support this and make the change to the `lastClaimed` value.

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -16,6 +17,8 @@ contract CompoundIntegration is InitializableAbstractIntegration {
     event RewardTokenCollected(address recipient, uint256 amount);
     event SkippedWithdrawal(address bAsset, uint256 amount);
 
+    uint256 internal lastClaimed;
+
```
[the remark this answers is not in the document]
shapeshed: Something like this 

```
      /**
       * @dev Approves Liquidator to spend reward tokens
       * @param _address Reward token
       */
      function approveRewardToken(
          address _rewardToken
      )
          external
          onlyGovernor
      {
          address liquidator = nexus.getModule(keccak256("Liquidator"));
          require(liquidator != address(0), "Liquidator address cannot be zero");
          MassetHelpers.safeInfiniteApprove(_rewardToken, liquidator);

          emit RewardTokenApproved(_rewardToken, liquidator);
      }

```
If the liquidator changes we can call this again and ask nexus for the new address. There is the issue that if the liquidator address changes then that contract still has approval. If that's important we could write the current liquidator address to storage, then compare what we get back from Nexus, revoking permission from the old address if it has changed. Or did you means something else by migration?

PR #115, opened 2020-11-12
Gas reductions + AaveV2

alsco77:
## Description

1. Major gas reductions in mUSD by adding a 'cache' as described in [MIP-3](https://mips.mstable.org/MIPS/mip-3)
2. Modifying interest collection mechanism to reduce SAVE deposit gas costs as described in [MIP-4](https://mips.mstable.org/MIPS/mip-4)
3. Adding a `PlatformIntegration` for Aave V2 and the ability to move assets between Integrations

**Affected files:**

 - `Masset.sol`: Major changes to depositing and withdrawing tokens (1) and tracking fee collection (2)
 - `BasketManager.sol`: Minor changes to accomodate (1) and (3), also ability to collect system revenue to recipient (to support buy and make)
 - `AaveIntegration.sol`, `CompoundIntegration.sol`: Major changes to accommodate new cache functionality (1) and maintain backwards compatibility
 - `SavingsManager.sol`: Extension of current streaming functionality to accomodate (2)

**New files:**

 - `AaveV2Integration.sol`: New contract

**Minor changes:**

 - Changed default `swapFee` and `redemptionFee` from `40bps` and `0bps` to `6bps` and `3bps` respectively, as this is what is currently active on the mainnet mUSD contract. Note these changes don't have any knock on effect to mainnet and are used to simulate live environment as accurately as possible
 - Lowered minimum `savingsRate` from `90%` to `70%`. Having a `savingsRate` of `70%` would mean that `70%` of system revenue is sent to SAVE and the other 30 is redistributed to a specified address. This is now more in line with general standards

## Upgrade plan

`Integration` files are all backwards compatible, as is the `BasketManager` and `SavingsManager`. Therefore it would be possible to upgrade all these contracts without completing (1), (2) or (3). The below plan iteratively moves towards full completion. 

1. Upgrade `AaveIntegration`, `CompoundIntegration` and `BasketManager`. Deploy `AaveV2Integration`. Complete (3) - migration of funds from `AaveIntegration` to `AaveV2Integration`
2. Upgrade `SavingsManager` (first, ensure that there is no residual mUSD from the Liquidator stream here)
3. Upgrade `Masset` and set cache size to finalise (1) and (2)

**Low level** 

- [x] Resolve remaining TODO's (removing console logs & review withdrawRaw)
  - [x] Account for AAVE yield farming distribution?
- [x] Deploy all contracts
- [ ] Propose upgrades
- [ ] Accept and complete step 1
- [ ] Accept and complete step 2
- [ ] Accept and complete step 3

### 1 - Cache

Details in [MIP-3](https://mips.mstable.org/MIPS/mip-3).

There are a number of concerns in accounting for tokens with transfer fees enabled. If we were to use the cache with these assets then it may be possible to churn funds at both a cost to the attacker and to the system. It is highly unlikely for this to happen, but the `require(transferred == _quantity, "Asset not fully transferred");` in `_depositTokens` should protect against this churn, and also ensure that when we deposit into the cache, we are minting the correct amount of mAsset.

Some properties:
 - if a given bAsset has `txFees` enabled (e.g. USDT turns on), then there should be 0 in the cache

### 2 - Swap fee accrual and interest collection batching

Details in [MIP-4](https://mips.mstable.org/MIPS/mip-4).

Adds a new `surplus` variable to the `Masset.sol`. This always remains at `1` base unit or higher to limit the cost to Swap/Redeem.
**Note**: the fees do not get minted until collected by the SAVE contract. Technically speaking, the sum of the vaultBalances should equal `totalSupply` + `surplus`.. so with that being the case, this is what is sent to the ForgeValidator to use in weight checking.

### 3 - Aave V2 and migration

Aave are releasing V2. Instead of upgrading the existing migration and then doing some sort of migration script, it's much less error prone to simply create a second platform integration contract and move the funds about.

Upgrades the Aave Integration to accommodate [Aave v2](https://medium.com/aave/aave-v2-the-seamless-finance-d52075d97a70) - [docs here](https://docs.aave.com/v2/-MJXUluJ2u1DiL-VU6MM/changes-from-v1-to-v2).

#### V2 Spec

V1 -> V2 changes:
 - New `LendingPoolAddressesProvider` that gives out all different module addresses
 - [LendingPool.zip](https://github.com/mstable/mStable-contracts/files/5535727/LendingPool.zip) interface changes: `deposit` adds an `_onBehalfOf` arg
 - `aToken` addresses change
 - LendingPoolCore (`ILendingPoolAddressesProvider(platformAddress).getLendingPoolCore();`) is no longer the contract to approve - now simply `LendingPool`
 - Redeem and Withdraw via LendingPool, not aToken

Assumptions: 
 - The AToken interface will stay the same
 - The AToken still passively increases in balance
 - The decimals on AToken will remain the same
 - The referral code will stay the same


contracts/masset/platform-integrations/AaveV2Integration.sol
```diff
@@ -0,0 +1,255 @@
...
+     *      (mAsset and corresponding BasketManager)
+     * @param _bAsset              Address for the bAsset
+     * @param _amount              Units of bAsset to deposit
+     * @param _isTokenFeeCharged   Flag that signals if an xfer fee is charged on bAsset
+     * @return quantityDeposited   Quantity of bAsset that entered the platform
+     */
+    function deposit(
+        address _bAsset,
+        uint256 _amount,
+        bool _isTokenFeeCharged
```
wuestholz, line 36: Maybe use `_hasTxFee` here for consistency.

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -115,23 +149,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+        address _receiver,
+        address _bAsset,
+        uint256 _amount
+    )
+        external
+        onlyWhitelisted
+        nonReentrant
+    {
+        // @TODO - add protections to ensure bAsset is logged to aToken?
+        // Unecessary only Masset/BM have write access
```
wuestholz, line 189: Unecessary => Unnecessary

contracts/masset/platform-integrations/AaveIntegration.sol
```diff
@@ -60,43 +59,104 @@ contract AaveIntegration is InitializableAbstractIntegration {
...
+        address _receiver,
+        address _bAsset,
+        uint256 _amount
+    )
+        external
+        onlyWhitelisted
+        nonReentrant
+    {
+        // @TODO - add protections to ensure bAsset is logged to aToken?
+        // Unecessary only Masset/BM have write access
```
wuestholz, line 152: Unecessary => Unnecessary

contracts/savings/SavingsManager.sol
```diff
@@ -118,6 +130,33 @@ contract SavingsManager is ISavingsManager, PausableModule {
         IERC20(_mAsset).safeApprove(address(_savingsContract), uint256(-1));
     }
 
+    /**
+     * @dev Freezes streaming of mAssets
+     */
+    function freezeStreams()
+        external
+        onlyGovernor
```
wuestholz, line 138: Maybe add `whenStreamsNotFrozen`.

contracts/masset/Masset.sol
```diff
@@ -1,10 +1,14 @@
 pragma solidity 0.5.16;
 pragma experimental ABIEncoderV2;
 
+// TODO - remove
```
wuestholz, line 4: I assume all this logging will be removed before merging.

alsco77: Yes, all `TODO`'s will be actioned and removed before deployment

contracts/masset/Masset.sol
```diff
@@ -510,20 +570,20 @@ contract Masset is
...
-        ForgePropsMulti memory props = basketManager.prepareForgeBassets(_bAssets, _bAssetQuantities, false);
+        RedeemProps memory props = basketManager.prepareRedeemBassets(_bAssets);
         if(!props.isValid) return 0;
 
+        Cache memory cache = _getCacheDetails();
+
         // Validate redemption
         (bool redemptionValid, string memory reason, bool applyFee) =
-            forgeValidator.validateRedemption(basket.failed, totalSupply(), basket.bassets, props.indexes, _bAssetQuantities);
+            forgeValidator.validateRedemption(false, cache.vaultBalanceSum, props.allBassets, props.indexes, _bAssetQuantities);
```
wuestholz, line 586: It seems like some of the validation now happens earlier (in `prepareRedeemBassets`). For instance, if `basket.failed` holds the following precondition would fail: https://github.com/mstable/mStable-contracts/blob/gas-reductions/contracts/masset/BasketManager.sol#L105. Previously, it seems like this would fail and it would be the validator's responsibility: https://github.com/mstable/mStable-contracts/blob/gas-reductions/contracts/masset/forge-validator/ForgeValidator.sol#L227. Is this correct and the intended behavior?

alsco77: Yes that is intentional - this single piece of validation was moved to the `prepareRedeemBassets`. The goal was to reduce the number of calls to BasketManager and consequently the quantity of data being returned.

wuestholz: I see. Thanks! Would it make sense to update the validator to eliminate the code that is probably dead now?

alsco77: It would, but I weighed up the effort of doing another contract deploy and upgrade against simply setting this flag to false, and decided to just set the flag here.

wuestholz: Ok, that makes sense. It might be good to keep track of such things in case you're redeploying contracts anyways.

contracts/savings/SavingsManager.sol
```diff
@@ -310,25 +414,32 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+
         IERC20 mAsset = IERC20(_mAsset);
         uint256 balance = mAsset.balanceOf(address(this));
+        uint256 leftover = _allUnclaimedRewards(_mAsset);
+
+        uint256 unallocated = balance.sub(leftover);
 
-        emit InterestWithdrawnByGovernor(_mAsset, _recipient, balance);
+        mAsset.approve(address(recipient), unallocated);
+        recipient.depositFunds(_mAsset, unallocated);
```
wuestholz, line 441: This isn't really depositing the funds per se. It's essentially notifying the recipient that it was approved, right?

contracts/savings/SavingsManager.sol
```diff
@@ -118,6 +130,34 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+
+        emit StreamsFrozen();
+    }
+
+    /**
+     * @dev Adds a new revenue recipient
+     * @param _mAsset           Address of underlying mAsset
+     * @param _recipient        Address of the recipient
+     */
+    function addRevenueRecipient(address _mAsset, address _recipient)
```
wuestholz, line 151: I suppose this could also remove a revenue recipient, right? If so, it might be worth considering to change the function name to `setRevenueRecipient` (similar for the event).

contracts/savings/SavingsManager.sol
alsco77: Yes - I should rename this to 'notifyRedistributionAmount'

contracts/savings/SavingsManager.sol
alsco77: Will change to `setRevenueRecipient`

contracts/masset/BasketManager.sol
```diff
@@ -239,8 +240,14 @@ contract BasketManager is
...
         for(uint8 i = 0; i < count; i++) {
             Basset memory b = allBassets[i];
-            // call each integration to `checkBalance`
-            uint256 balance = IPlatformIntegration(integrations[i]).checkBalance(b.addr);
+            address bAsset = b.addr;
+
+            // call each integration to `checkBalance` and sum with the cache balance
+            address integration = integrations[i];
+            uint256 lending = IPlatformIntegration(integration).checkBalance(bAsset);
+            uint256 cache = IERC20(bAsset).balanceOf(integration);
```
wuestholz, line 248: Consider changing this line to something like this to account for assets that might be transferred to the integration:
```
uint256 cache = 0;
if (!b.isTransferFeeCharged) {
  cache = IERC20(bAsset).balanceOf(integration);
}
```

contracts/masset/shared/MassetHelpers.sol
```diff
@@ -18,25 +18,20 @@ library MassetHelpers {
...
-            IERC20(_basset).safeTransferFrom(_sender, _recipient, _qty);
-            uint256 balAfter = IERC20(_basset).balanceOf(_recipient);
-            receivedQty = StableMath.min(_qty, balAfter.sub(balBefore));
-        } else {
-            IERC20(_basset).safeTransferFrom(_sender, _recipient, _qty);
-        }
+        uint256 balBefore = IERC20(_basset).balanceOf(_recipient);
+        IERC20(_basset).safeTransferFrom(_sender, _recipient, _qty);
+        recipientBalance = IERC20(_basset).balanceOf(_recipient);
+        receivedQty = StableMath.min(_qty, recipientBalance.sub(balBefore));
```
wuestholz, line 34: I might be worth considering that this function may fail if `recipientBalance < balBefore`. I suppose this can, for instance, happen if _sender and _recipient are identical and there is a transfer fee.

contracts/masset/Masset.sol
```diff
@@ -656,10 +832,48 @@ contract Masset is
...
+      * @param _cacheSize Maximum percent of total mAsset supply to hold for each bAsset
+      */
+    function setCacheSize(uint256 _cacheSize)
+        external
+        onlyGovernance
+    {
+        require(_cacheSize <= 2e17, "Must be <= 20%");
+
+        cacheSize = _cacheSize;
+
```
wuestholz, line 872: A reduction in cache size might temporily violate the invariant that the cached value is <= cacheSize * totalSupply. Is this intended? Or should it trigger a deposit into the lending pool?

contracts/masset/Masset.sol
```diff
@@ -249,41 +266,65 @@ contract Masset is
...
-        quantityDeposited = StableMath.min(deposited, _quantity);
-    }
+            quantityDeposited = transferred;
+
+            uint256 relativeMaxCache = _maxCache.divRatioPrecisely(_bAssetRatio);
 
+            console.log("_depositTokens: cacheBal: %s vs relativeMaxCache: %s", cacheBal, relativeMaxCache);
+            if(cacheBal >= relativeMaxCache){
+                uint256 delta = cacheBal.sub(relativeMaxCache.div(2));
+                IPlatformIntegration(_integrator).deposit(_bAsset, delta, false);
```
wuestholz, line 311: Maybe this should only deposit if `0 < delta` since the integration will fail otherwise. Maybe `cacheBal >= relativeMaxCache` could be changed to `cacheBal > relativeMaxCache`.

contracts/masset/Masset.sol
```diff
@@ -304,44 +345,61 @@ contract Masset is
...
+                recipient: recipient,
+                ratio: outputDetails.bAsset.ratio,
+                maxCache: cache.maxCache,
+                vaultBalance: outputDetails.bAsset.vaultBalance
+            })
+        );
+
+        basketManager.decreaseVaultBalance(outputDetails.index, outputDetails.integrator, amt.net);
+
+        surplus = cache.surplus.add(amt.scaledFee);
```
wuestholz, line 400: I don't quite understand why `surplus` is part of the cache struct. Wouldn't `cache.surplus` be equal to `surplus` here?

contracts/masset/Masset.sol
```diff
@@ -304,44 +345,61 @@ contract Masset is
...
+
+        emit Swapped(msg.sender, args.input, args.output, amt.net, recipient);
```
wuestholz, line 402: It seems like `output` wasn't assigned here.

contracts/masset/Masset.sol
```diff
@@ -573,46 +644,138 @@ contract Masset is
...
+            // 2. If txFee then short circuit - there is no cache
+            if(args.hasTxFee){
+                console.log("_withdrawTokens: hasTxFee");
+                IPlatformIntegration(args.integrator).withdraw(args.recipient, args.bAsset, amount.net, amount.net, true);
+            }
+            // 3. Else, withdraw from either cache or main vault
+            else {
+                uint256 cacheBal = IERC20(args.bAsset).balanceOf(args.integrator);
+                // 3.1 - If balance b in cache, simply withdraw
+                if(cacheBal > amount.net) {
```
wuestholz, line 758: Why not `cacheBal >= amount.net`?

contracts/masset/Masset.sol
```diff
@@ -573,46 +644,138 @@ contract Masset is
...
+                    console.log("_withdrawTokens: cacheBal > net - '%s' > '%s'", cacheBal, amount.net);
+                    IPlatformIntegration(args.integrator).withdrawRaw(args.recipient, args.bAsset, amount.net);
+                }
+                // 3.2 - Else reset the cache to X, or as far as possible
+                //       - Withdraw X+b from platform
+                //       - Send b to user
+                else {
+                    console.log("_withdrawTokens: cacheBal < net - '%s' < '%s'", cacheBal, amount.net);
+                    uint256 relativeMidCache = args.maxCache.divRatioPrecisely(args.ratio).div(2);
+                    uint256 totalWithdrawal = StableMath.min(relativeMidCache.sub(cacheBal).add(amount.net), args.vaultBalance.sub(cacheBal));
```
wuestholz, line 768: Is it possible for `relativeMidCache` to be less than `cacheBal`? Seems like it might be possible if the cache size was decreased earlier. It would be good to make sure that none of the safemath operations revert (maybe by changing the order).

Another question: What's the reason for considering `args.vaultBalance.sub(cacheBal)`? It seems like `vaultBalance` might be less than `cacheBal` if someone transferred tokens to the cache.

contracts/masset/Masset.sol
```diff
@@ -86,7 +95,9 @@ contract Masset is
         basketManager = IBasketManager(_basketManager);
 
         MAX_FEE = 2e16;
-        swapFee = 4e15;
+        swapFee = 6e14;
+        redemptionFee = 3e14;
+        cacheSize = 1e17;
```
wuestholz, line 100: It might be good to test extreme values here: for instance, by running the test suite with 0, 1, and a value that corresponds to 100%.

contracts/masset/Masset.sol
```diff
@@ -729,21 +943,46 @@ contract Masset is
...
+    function collectInterest()
+        external
+        onlySavingsManager
+        nonReentrant
+        returns (uint256 swapFeesGained, uint256 newSupply)
+    {
+        uint256 toMint = 0;
+        if(surplus > 1){
+            toMint = surplus.sub(1);
+            surplus = 1;
```
wuestholz, line 960: It might be good to add a comment here to explain why surplus isn't set to 0 (see https://mips.mstable.org/MIPS/mip-4). This seems to retain a tiny amount of fees in the contract. However, the "stuck" amount is negligible.

contracts/masset/platform-integrations/CompoundIntegration.sol
```diff
@@ -115,23 +149,50 @@ contract CompoundIntegration is InitializableAbstractIntegration {
...
+        uint256 _amount
+    )
+        external
+        onlyWhitelisted
+        nonReentrant
+    {
+        // @TODO - add protections to ensure bAsset is logged to aToken?
+        // Unnecessary only Masset/BM have write access
+
+        require(_amount > 0, "Must withdraw something");
```
wuestholz, line 191: It might make sense to validate that `_receiver != address(0)`. I don't think all tokens check this.

contracts/savings/SavingsManager.sol
```diff
@@ -118,6 +130,34 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
 
+    /**
+     * @dev Freezes streaming of mAssets
+     */
+    function freezeStreams()
+        external
+        onlyGovernor
+        whenStreamsNotFrozen
+    {
+        streamsFrozen = true;
```
wuestholz, line 141: Should it be possible to also "unfreeze" a stream?

contracts/masset/Masset.sol
```diff
@@ -656,10 +832,48 @@ contract Masset is
...
     ****************************************/
 
+    /**
+      * @dev Sets the MAX cache size for each bAsset. The cache will actually revolve around
+      *      _cacheSize * totalSupply / 2 under normal circumstances.
+      * @param _cacheSize Maximum percent of total mAsset supply to hold for each bAsset
+      */
+    function setCacheSize(uint256 _cacheSize)
+        external
+        onlyGovernance
```
wuestholz, line 867: Other functions, such as `setSwapFee` are more restrictive in who can call them (i.e., `onlyGovernor`). Is there a specific reason for this?

contracts/savings/SavingsManager.sol
```diff
@@ -141,31 +181,95 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+            uint256 leftover = _allUnclaimedRewards(_mAsset);
+
+            // Bundle this yield in with liquidator yield
+            // If (remainingTime < 24h)
+            //       take all remaining, add this, stream over next 24h
+            // else stream over remainingTime
+            uint256 end = streamEnd[_mAsset];
+            uint256 remaining = end > currentTime ? end.sub(currentTime) : 0;
+            uint256 newDuration = remaining < ONE_DAY ? ONE_DAY : remaining;
+            _initialiseStream(_mAsset, interestCollected.add(leftover), newDuration);
```
wuestholz, line 231: In MIP-4 there are two different (contradictory?) durations: 1 week and 1 day. I assume 1 day is the correct one.

contracts/masset/Masset.sol
alsco77: The next time an action (mint,swap,redeem) is called, it should even itself out based on the new cache size.

contracts/masset/Masset.sol
alsco77: Will add.

contracts/savings/SavingsManager.sol
alsco77: The freezing was just for doing a migration from one SavingsManager to the next.. however now I will add in a 'whenNotPaused' instead.

contracts/masset/Masset.sol
alsco77: The mStable community DAO also has write access to some of the parameters, notably things like the cache size. These will be phased over gradually to onlyGovernance

contracts/savings/SavingsManager.sol
alsco77: 1 day, yes. I will update the MIP

contracts/masset/shared/MassetHelpers.sol
alsco77: Given that we trust the caller here - I see no reason why one caller would ever want to transfer tokens to itself and no reason to account for this scenario

contracts/masset/Masset.sol
alsco77: I don't see a scenario in which the delta is == 0

contracts/masset/Masset.sol
alsco77: Yes but storing it in the cache struct avoid a second SLOAD?

contracts/masset/Masset.sol
alsco77: Sorry I don't understand

contracts/masset/Masset.sol
alsco77: No reason why not. Just updated this.

contracts/masset/Masset.sol
wuestholz: Hm, where would the two SLOADs be? It seems like this is the only place where `surplus` would be read and updated. But maybe I'm missing something.

contracts/masset/Masset.sol
wuestholz: It seems like the return value `output` is never assigned along this execution path. It would just be 0 always.

contracts/masset/Masset.sol
alsco77: Fixed the LHS by swapping the add and subtract around, since amount.net is > cacheBal.

RHS - vaultBalance.sub(cacheBal) must be considered because the total withdrawal can at maximum be everything that is in the lending market (this is the vaultbalance minus what is in the cache and does not include any interest that has been generated).

If the cacheBal is > vaultBalance, we simply cannot reach this point, and so this can be left as it is. Why? Because if the previous statement 'if(cacheBal >= amount.net)'. If amount.net was greater than the cache balance, and the cache was greater than the vault balance, then the amount.net is greater than the vault balance... which would be picked up in the Forge Validator

contracts/masset/Masset.sol
wuestholz: Yeah, I think the only scenario I see right now is if both `cacheBal` and `relativeMaxCache` are 0. I think `cacheBal` should always be positive though.

I think the change I'm proposing would simplify the reasoning here. First, it should be easy to see that `delta` is always positive. Second, to me it's more intuitive to think of `cacheBal == relativeMaxCache` as a valid state of the cache since the balance hasn't exceeded the maximum.

contracts/masset/BasketManager.sol
alsco77: You are right. Done.

contracts/masset/shared/MassetHelpers.sol
wuestholz: Yeah, you can always look at the callers to make sure this doesn't happen. For me, it simplifies reasoning about library functions if they make as few assumptions as possible. However, you're right that for its current callers this shouldn't be an issue.

contracts/masset/Masset.sol
wuestholz: Yeah, I agree. This observation is mostly to make sure you're aware that there may be short periods of time where the invariant does not hold. This is mostly important if you're assuming the invariant in other places of the code.

contracts/savings/SavingsManager.sol
wuestholz: I see. Thanks!

contracts/masset/Masset.sol
wuestholz: I see. I assumed something like this might be the reason, but wanted to make sure it wasn't a typo. :) Thanks for clarifying!

contracts/masset/Masset.sol
wuestholz: Ah, I think I see it now. You also read it to compute `vaultBalanceSum`.

contracts/masset/Masset.sol
alsco77: Gotcha, resolved.

contracts/masset/Masset.sol
alsco77: Yes :) Please resolve the comment here if you see no further issues with that. Note that the sum of the vault balances should be equal to surplus + totalSupply.

contracts/masset/Masset.sol
wuestholz: Thanks!

About the LHS: Actually, then why not even change the precedence from (relativeMidCache + amount.net) - cacheBal to relativeMidCache + (amount.net - cacheBal)? That seems even slightly more easy to reason about.

About the RHS: I see. Yeah, I think this makes sense. I wonder if you can use a similar argument to show that LHS <= RHS, thus making the min-operation unnecessary.

contracts/masset/Masset.sol
wuestholz: I don't seem to be able to mark it as resolved, but please do so. :)

contracts/savings/SavingsManager.sol
```diff
@@ -140,32 +180,98 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+     * @param _mAsset The mAsset in question
+     * @return leftover The total amount of mAsset that is yet to be collected from a stream
+     */
+    function _allUnclaimedRewards(address _mAsset) internal view returns (uint256 leftover) {
+        uint256 currentTime = now;
+
+        uint256 end = streamEnd[_mAsset];
         uint256 lastUpdate = lastCollection[_mAsset];
         uint256 unclaimedSeconds = 0;
         if(currentTime <= end || lastUpdate < end){
```
wuestholz, line 253: What's the reason for the disjunct `currentTime <= end`?

alsco77: Good catch.. this may have been an oversight from when i took inspiration from line 373. Current time can't be < lastUpdate, therefore this is redundant.

contracts/savings/SavingsManager.sol
```diff
@@ -140,32 +180,98 @@ contract SavingsManager is ISavingsManager, PausableModule {
...
+            // Get remaining rewards
+            uint256 leftover = _allUnclaimedRewards(_mAsset);
+
+            // Bundle this yield in with liquidator yield
+            // If (remainingTime < 24h)
+            //       take all remaining, add this, stream over next 24h
+            // else stream over remainingTime
+            uint256 end = streamEnd[_mAsset];
+            uint256 remaining = end > currentTime ? end.sub(currentTime) : 0;
+            uint256 newDuration = remaining < ONE_DAY ? ONE_DAY : remaining;
```
wuestholz, line 232: It seems like there are two different durations that share the same stream: 1 day and 1 week. This seems to complicate determining the new duration slightly. For instance, there might be 6 days remaining. Now, when a 1 day stream is added the yield is bundled with the existing one and streamed over a longer period.

Have you considered using to separate streams?

contracts/savings/SavingsManager.sol
```diff
@@ -231,12 +337,12 @@ contract SavingsManager is ISavingsManager, PausableModule {
 
             // 4. Distribute the interest
             //    Calculate the share for savers (95e16 or 95%)
-            uint256 saversShare = interestCollected.mulTruncate(savingsRate);
+            uint256 saversShare = interestCollected.add(newReward).mulTruncate(savingsRate);
```
wuestholz, line 340: Does this fix a minor issue due to previously not truncating the new reward?

contracts/masset/platform-integrations/AaveV2Integration.sol
```diff
@@ -0,0 +1,256 @@
...
+        }
+
+        emit Deposit(_bAsset, address(aToken), quantityDeposited);
+    }
+
+    /**
+     * @dev Withdraw a quantity of bAsset from the platform
+     * @param _receiver     Address to which the bAsset should be sent
+     * @param _bAsset       Address of the bAsset
+     * @param _amount       Units of bAsset to withdraw
```
wuestholz, line 66: Add documentation for `_hasTxFee`.

contracts/masset/BasketManager.sol
```diff
@@ -485,6 +503,68 @@ contract BasketManager is
...
+    {
+        uint256 len = _bAssets.length;
+        require(len > 0, "Must migrate some bAssets");
+
+        for(uint i = 0; i < len; i++){
+            // 1. Check that the bAsset is in the basket
+            address bAsset = _bAssets[i];
+            (bool inBasket, uint8 index) = _isAssetInBasket(bAsset);
+            require(inBasket, "bAsset does not exist");
+
```
wuestholz, line 528: Maybe it would make sense to add something like `require(!basket.bassets[index].isTransferFeeCharged` here.

contracts/masset/BasketManager.sol
```diff
@@ -485,6 +503,68 @@ contract BasketManager is
...
+            // 2. Withdraw everything from the old platform integration
+            IPlatformIntegration oldIntegration = IPlatformIntegration(integrations[index]);
+            require(address(oldIntegration) != _newIntegration, "Must transfer to new integrator");
+            // 2.1. Withdraw from the lending market
+            uint256 lendingBal = oldIntegration.checkBalance(bAsset);
+            uint256 cache = IERC20(bAsset).balanceOf(address(oldIntegration));
```
wuestholz, line 534: I would move this line to before the call to `withdrawRaw`.

contracts/masset/BasketManager.sol
```diff
@@ -485,6 +503,68 @@ contract BasketManager is
...
+
+            // 3. Update the integration address for this bAsset
+            integrations[index] = _newIntegration;
+
+            // 4. Deposit everything into the new
+            //    This should fail if we did not receive the full amount from the platform withdrawal
+            // 4.1. Deposit all bAsset
+            IERC20(bAsset).safeTransfer(_newIntegration, total);
+            IPlatformIntegration newIntegration = IPlatformIntegration(_newIntegration);
+            newIntegration.deposit(bAsset, lendingBal, false);
```
wuestholz, line 548: It might be good to check here that the return value matches `lendingBal`.

contracts/masset/BasketManager.sol
```diff
@@ -485,6 +503,68 @@ contract BasketManager is
...
+            // 4.2. Check balances
+            uint256 newLendingBal = newIntegration.checkBalance(bAsset);
+            uint256 newCache = IERC20(bAsset).balanceOf(address(newIntegration));
+            uint256 upperMargin = 1001e15;
```
wuestholz, line 552: What's the reason for enforcing an upper bound?

contracts/masset/BasketManager.sol
```diff
@@ -485,6 +503,68 @@ contract BasketManager is
...
+            uint256 lowerMargin =  999e15;
+
+            require(
+                newLendingBal >= lendingBal.mulTruncate(lowerMargin) &&
+                newLendingBal <= lendingBal.mulTruncate(upperMargin),
+                "Must transfer full amount");
+            require(
+                newCache >= cache.mulTruncate(lowerMargin) &&
```
wuestholz, line 560: Couldn't the lower bound be more strict (e.g., `cache`) given that the token does not charge TX fees?

contracts/masset/Masset.sol
alsco77: Agreed, changed.

contracts/masset/Masset.sol
alsco77: LHS - agree on the rationale, updated.

RHS - the min operation **is** necessary.. LHS > RHS in the majority of cases.

contracts/savings/SavingsManager.sol
alsco77: Yes, "streamed" rewards (via liquidator) were not being taken into account for transferring to governance.

contracts/masset/BasketManager.sol
alsco77: Incase for some reason the new integrations `checkBalance` call returns some number with completely different precision.. it makes sense to just check here in case we cause hyper inflation.

contracts/masset/BasketManager.sol
alsco77: Have added an extra decimal of precision on.

contracts/savings/SavingsManager.sol
alsco77: This was done primarily to reduce gas costs as much as possible. The goal was to share the `lastCollection` property.

I will action this by splitting out into two streams, and if possible, continuing to share the `lastCollection`. If not, then I will introduce a second timestamp for the platform interest yield.

Sharing the same collection time = 2x SLOAD 
Having a separate collection time. = 3x SLOAD and 1x SSTORE

alsco77: Created two seperate streams, both calling _collectAndDistributeInterest before they are set. This causes the lastCollected timestamp to always be set correctly.

PR #116, opened 2020-11-14
docs: add note about contract artifacts in README

PaulRBerg:
It's not clear that the contract artifacts are meant to be used via the `@mstable/protocol` npm package. This PR adds an explicit note for this.


alsco77 (approved):
Thanks Paul!

PR #119, opened 2020-12-09
Add boost formula

osolmaz:


alsco77:
Merging in to branch save-v2, will add test suite for it there

PR #159, opened 2021-04-03
Flexible SaveWrapper

JamesLefrere:
- Make SaveWrapper stateless (that is, aside from being `Ownable`) so it can be flexible about assets supported
- Define pattern for discrete Hardhat tasks with params
- Disable some eslint rules

Tested on Ropsten, including a test sBTC/WETH pair on Uniswap.

### DIFF
--- a/contracts/savings/peripheral/SaveWrapper.sol
+++ b/contracts/savings/peripheral/SaveWrapper.sol
@@ -6,9 +6,10 @@ import { IMasset } from "../../interfaces/IMasset.sol";
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
 import { IUniswapV2Router02 } from "../../interfaces/IUniswapV2Router02.sol";
 import { IBasicToken } from "../../shared/IBasicToken.sol";
-
 
 interface IBoostedSavingsVault {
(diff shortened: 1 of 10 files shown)


tasks/deployAaveIntegration.ts
```diff
@@ -30,6 +30,8 @@ task("deployAaveIntegration", "Deploys an instance of AaveV2Integration contract
         addresses.nexus,
         addresses.mAsset,
         addresses.aave,
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+        // @ts-ignore
```
JamesLefrere, line 34: I guess the ABI has changed here

contracts/savings/peripheral/SaveWrapper.sol
```diff
@@ -18,125 +19,205 @@ interface IBoostedSavingsVault {
...
+     * @param _amountOutMin   Min uniswap output in bAsset units
+     * @param _path           Sell path on Uniswap (e.g. [WETH, DAI])
+     * @param _minOutMStable  Min amount of mAsset to receive
+     * @param _stake          Add the imAsset to the Savings Vault?
      */
     function saveViaUniswapETH(
+        address _mAsset,
+        address _save,
+        address _vault,
+        address _uniswap,
```
JamesLefrere, line 99: Theoretically this could work with SushiSwap, I think.

alsco77 (approved):
Nice work

contracts/savings/peripheral/SaveWrapper.sol
alsco77: Yep

PR #167, opened 2021-04-18
Migrate contracts & add husky precommits

alsco77:
- Migrates SavingsManager (and test cases)
 - Moves duplicated interest validation logic to library
 - Adds husky precommit validation


contracts/savings/SavingsManager.sol
```diff
@@ -0,0 +1,432 @@
...
+                );
+
+            // Get remaining rewards
+            uint256 leftover = _unstreamedRewards(_mAsset, StreamType.yield);
+            _initialiseStream(_mAsset, StreamType.yield, interestCollected + leftover, ONE_DAY);
+
+            emit InterestCollected(_mAsset, interestCollected, totalSupply, apy);
+        }
+
+        emit InterestCollected(_mAsset, interestCollected, totalSupply, 0);
```
naddison36, line 236: I think this needs to be in an `else` from `if (interestCollected > 0)`. At the moment the interest collection transactions are firing two `InterestCollected` events

PR #169, opened 2021-04-21
Restructure Masset inheritance

alsco77:


contracts/interfaces/IInvariantValidator.sol
```diff
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-or-later
 pragma solidity 0.8.2;
 
-import { BassetData, InvariantConfig } from "../masset/MassetStructs.sol";
+import "../masset/MassetStructs.sol";
```
naddison36, line 4: Any reason not to use explicit imports? I prefer implicit imports as it makes it easier to trace where contracts, interfaces, structs etc come from. But it doesn't really matter so no need to change.

contracts/masset/Masset.sol
```diff
@@ -2,21 +2,21 @@
 pragma solidity 0.8.2;
 pragma abicoder v2;
 
-// External
-import { IInvariantValidator } from "../interfaces/IInvariantValidator.sol";
+import "hardhat/console.sol";
```
naddison36, line 5: This should be commented out for mainnet

contracts/masset/Masset.sol
```diff
@@ -227,7 +203,22 @@ contract Masset is
         uint256 _minOutputQuantity,
         address _recipient
     ) external override nonReentrant whenHealthy returns (uint256 mintOutput) {
-        mintOutput = _mintTo(_input, _inputQuantity, _minOutputQuantity, _recipient);
+        require(_recipient != address(0), "Invalid recipient");
```
naddison36, line 206: Are we not using an internal `_mintTo` function to reduce the contract size? The last time we looked at this calling an internal function used less gas which is counter intuitive.

contracts/masset/Masset.sol
```diff
@@ -787,39 +560,39 @@ contract Masset is
...
      */
     function getBassets()
         external
         view
         override
-        returns (BassetPersonal[] memory personal, BassetData[] memory data)
+        returns (BassetPersonal[] memory personal, BassetData[] memory bData)
     {
-        return (bAssetPersonal, bAssetData);
+        return (data.bAssetPersonal, data.bAssetData);
```
naddison36, line 577: If this was called in a transaction, it could be more gas efficient to have
```
personal = data.bAssetPersonal;
bData = data.bAssetData;
```
But this will mostly be static called so probably doesn't matter.

contracts/masset/MassetStructs.sol
```diff
@@ -73,6 +80,18 @@ struct FeederData {
     WeightLimits weightLimits;
 }
 
+struct MassetData {
+    uint256 swapFee;
+    uint256 redemptionFee;
```
naddison36, line 85: No `govFee` like the Feeder Pools?

naddison36 (approved):
Looks good to me.

contracts/interfaces/IInvariantValidator.sol
alsco77: `MassetStructs` for me is an exception given that there are so many individual structs there

contracts/masset/Masset.sol
alsco77: I agree with the counter intuitive point which was the rationale for keeping it inline

contracts/masset/Masset.sol
alsco77: Yeah this is only called externally

contracts/masset/MassetStructs.sol
alsco77: `govFee` is essentially equivalent to the `savingsRate` in SavingsManager, which applies to mAsset revenue

PR #175, opened 2021-05-13
Correct the misleading comment

tanliwei:
The result of `9e38 / 1e18` should be `9e20` instead of `9e18`.


tanliwei:
The more reasonable change/PR may be https://github.com/mstable/mStable-contracts/pull/176

PR #176, opened 2021-05-13
Correct the misleading comment

tanliwei:
The result of "9e38 / 1e18" should be "9e20" instead of "9e18".
I guess, what the comment intends to explain is "9e36 / 1e18 = 9e18", based on the context.


alsco77 (approved):
Thanks!

PR #178, opened 2021-05-18
Added Hardhat task to poke boosted vaults

naddison36:
- [x] Added Poker contract to bulk poke accounts across multiple boosted vaults
- [x] Added graphql, graphql-multicall and ethers-multicall dependencies
- [x] Added `over-boost` Hardhat task with `update` option flag
- [x] Added `deployPoker` Hardhat task that deploys the new `Poker` contract
- [x] Moved `impersonate` function used in mainnet fork tests to `@utils/fork`


tasks/poker.ts
```diff
@@ -0,0 +1,174 @@
...
+            )
+            console.log("Account, Raw Balance, Boosted Balance, vMTA balance, vMTA diff, Boost Actual, Boost Expected, Boost Diff")
+            vault.accounts.forEach((account, i) => {
+                const boostActual = BN.from(account.boostedBalance).mul(1000).div(account.rawBalance).toNumber()
+                const boostExpected = calcBoost(BN.from(account.rawBalance), vMtaBalancesMap[account.account.id], priceCoeff, boostCoeff)
+                    .div(BN.from(10).pow(15))
+                    .toNumber()
+                const boostDiff = boostActual - boostExpected
+                // Calculate how much extra MTA is being distributed = vMTA balance * (actual boost - expected boost)
+                const vMtaExtra = vMtaBalancesMap[account.account.id].mul(boostDiff).div(1000)
```
alsco77, line 127: Actually I think this needs to be fixed. Imagine a users vMTA balance is now 0 - they could have a 3x boost on $1m and it would be filtered out here.

It makes much more sense to detect the value of a users boosted holdings - you can just do `account.boostedBalance * priceCoeff / 1e18` to find the USD value of their boost. I would say, if `boostDiff > 200 && USD balance > 50k` then poke.

alsco77 (changes requested):

PR #179, opened 2021-06-03
Liquidator of stkAAVE

naddison36:
- [x] claim stkAAVE from integration contracts
- [x] start colldown of stkAAVE
- [x] redeem AAVE from stkAAVE after cooldown
- [x] swap AAVE for bAsset. eg USDC, WBTC or GUSD
- [x] for mAssets, mint using bAsset and deposit into Saving Manager
- [x] for Feeder Pools, transfer bAsset to integration contract
- [x] Remove used of Curve from liquidator
- [x] Upgrade liquidator to use Uniswap V3
- [x] add reporting of accrued COMP and stkAAVE to snap tasks


contracts/masset/peripheral/CompoundInegration.sol
```diff
@@ -5,7 +5,7 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { MassetHelpers } from "../../shared/MassetHelpers.sol";
 import { AbstractIntegration } from "./AbstractIntegration.sol";
-import { ICERC20 } from "./ICompound.sol";
+import { ICERC20 } from "../../peripheral/Compound/ICompound.sol";
 
 /**
  * @title   CompoundIntegration
```
alsco77, line 11: File name is incorrect

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -17,119 +22,193 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
  * @author  mStable
  * @notice  The Liquidator allows rewards to be swapped for another token
  *          and returned to a calling contract
- * @dev     VERSION: 1.2
- *          DATE:    2020-12-16
+ * @dev     VERSION: 1.3
+ *          DATE:    2021-05-28
  */
-contract Liquidator is ILiquidator, Initializable, ImmutableModule {
+contract Liquidator is Initializable, ModuleKeysStorage, ImmutableModule {
```
alsco77, line 28: No interface applied here. Not a big deal since all are external calls

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -219,57 +304,207 @@ contract Liquidator is ILiquidator, Initializable, ImmutableModule {
...
+            // else the current time is past the unstake window so claim more stkAave and reactivate the cool down
+        }
+
+        // 1. For each Aave integration contract
+        uint256 len = aaveIntegrations.length;
+        for (uint256 i = 0; i < len; i++) {
+            address integrationAdddress = aaveIntegrations[i];
+
+            // 2. Claim the platform rewards on the integration contract. eg stkAave
+            PAaveIntegration(integrationAdddress).claimRewards();
```
alsco77, line 385: Should probably refactor this to be an interface as discussed

alsco77 (changes requested):
Wow, lots of code here! Generally looks good :+1:  However I think this Liquidator is fundamentally not scalable/overly complex. In hindsight, each integration/token should likely have its own liquidation mechanism.

I don't see why we shouldn't make these small QOL changes before rebasing.

contracts/masset/liquidator/Liquidator.sol
naddison36: The Liquidator interfaces are now different between the mainnet and Polygon chains so I updated the `contracts/masset/liquidator/ILiquidator.sol` interface to match the new mainnet Liquidator and add a new `contracts/polygon/IPLiquidator.sol` interface for Polygon.

contracts/masset/liquidator/Liquidator.sol
naddison36: I started adding `claimRewards` to `IPlatformIntegration` but that caused too many problems with the older Compound and Aave V2 integrations. I ended up adding a new `IClaimRewards` interface to remove the dependency to `PAaveIntegration` which causes issues with different `Initializable` contracts in the flattener.

contracts/masset/peripheral/CompoundInegration.sol
naddison36: done

alsco77 (approved):

PR #185, opened 2021-06-25
Pool Together changes

naddison36:
Changes so mStable contracts and functions can be imported into external repos. eg `pooltogether-mstable`

- [x] Added `underlying()` function to `ISavingsContractV2` interface used by the PoolTogether yield source contract
- [x] Updated from legacy `hardhat-typechain` to `@typechain/hardhat`
- [x] Upgraded to latest dependencies including ethers, typechain, hardhat and typescript
- [x] Export functions and types from the package
- [x] Added type definitions to TS config
- [x] moved typescript output from `transpiled` to `dist` and added to repo and published package
- [x] Fixed bug in `deployMV3` task that was causing type errors with new Typechain version
- [x] `Token` interface renamed optional `savings` to `vault` and added optional `saving`
- [x] Added `rewards` task to get `COMP` and `stkAAVE` stats


alsco77 (changes requested):
Solidity looks good

Can you give me a TLDR of the purpose of the dist folder here? If it is for cross repo sharing, we should just run a command during the `prepublishOnly` script that will add the necessary resources to the dist folder before publish, and then add the files to the `files` in package json.

What do you think?

PR #191, opened 2021-07-07
feat: Added BoostedDualVault and refactored file layout

alsco77:
To review:
- [ ] Compare BoostedDualVault diff vs BoostedVault
- [ ] Compare BoostedDualVault diff vs StakingRewardsWithPlatformToken
- [ ] Review logic
- [ ] Review tests
- [ ] Ensure file is deployable: there may be issues with bytecode limits


contracts/rewards/boosted-staking/BoostedDualVault.sol
```diff
@@ -0,0 +1,705 @@
...
+    /// @notice token the rewards are distributed in. eg MTA
+    IERC20 public immutable rewardsToken;
+    /// @notice token the platform rewards are distributed in. eg WMATIC
+    IERC20 public immutable platformToken;
+    /// @notice contract that holds the platform tokens
+    PlatformTokenVendor public platformTokenVendor;
+    /// @notice total raw balance
+    uint256 public totalRaw;
+
+    uint64 public constant DURATION = 7 days;
```
naddison36, line 54: Let's make `DURATION` immutable and set in the constructor so we have some flexibility on deployment on the period lengths

contracts/rewards/boosted-staking/BoostedDualVault.sol
```diff
@@ -0,0 +1,705 @@
...
+    ****************************************/
+
+    /**
+     * @dev Stakes a given amount of the StakingToken for the sender
+     * @param _amount Units of StakingToken
+     */
+    function stake(uint256 _amount)
+        external
+        override
+        updateReward(msg.sender)
```
naddison36, line 218: Why not have the `updateReward` and `updateBoost` modifiers on the internal `_stake` function passing the `beneficiary`? This is what the non-boosted `StakingRewards` contract does.

naddison36 (commented):
Looks good. I've just updated some of the Natspec.

I can also change the `DURATION` to be an immutable set on deployment. That gives us a little more flexibility

contracts/rewards/boosted-staking/BoostedDualVault.sol
alsco77: Personally I would rather keep all reward contracts with the same distribution, this should make automating them easier in the future, as everything will happen on a weekly cadence. I think it will be easier to manage. Can certainly make it an immutable and use it if there is a strong reason to in future

contracts/rewards/boosted-staking/BoostedDualVault.sol
alsco77: Hmm, not sure. I think I just wanted to keep all external fns with the modifiers to avoid getting confused and potentially missing something. The updateBoost has to be called at the end of the external fn anyway, i think it may introduce a risk if there is logic in the external fn, to move it to internal

naddison36: Fair enough. I must admit I got confused reading the `StakingRewards` contract with the modified on the internal functions.

contracts/rewards/boosted-staking/BoostedDualVault.sol
naddison36: ok, let's leave it as a constant. No need to change to an immutable.

naddison36 (approved):

PR #196, opened 2021-07-14
Liquidator upgrade for ALCX

naddison36:
- [x] Liquidator upgraded to liquidate ALCX from the Alchemix integration contract
  - [x] ALCX is sold for alUSD using Uniswap V3 with path ALCX > ETH > DAI > alUSD
  - [x] The alUSD is transferred to the integration contract
- [x] `rewards` task upgraded to report ALCX
- [x] `prepublishOnly` script to copy contract type definitions to `/dist/types/generated`
- [x] Feeder tasks add for redeem and swap
- [x] Token tasks added for approve

### DIFF
--- a/contracts/masset/liquidator/Liquidator.sol
+++ b/contracts/masset/liquidator/Liquidator.sol
@@ -350,27 +350,27 @@ contract Liquidator is ILiquidator, Initializable, ModuleKeysStorage, ImmutableM
             );
         uniswapRouter.exactInput(param);
 
-        // address mAsset = liquidation.mAsset;
-        // // If the integration contract is connected to a mAsset like mUSD or mBTC
-        // if (mAsset != address(0)) {
-        //     // 4a. Mint mAsset using purchased bAsset
-        //     uint256 minted = _mint(bAsset, mAsset);
+        address mAsset = liquidation.mAsset;
+        // If the integration contract is connected to a mAsset like mUSD or mBTC
+        if (mAsset != address(0)) {
+            // 4a. Mint mAsset using purchased bAsset
+            uint256 minted = _mint(bAsset, mAsset);
 
-        //     // 5a. Send to SavingsManager
-        //     address savings = _savingsManager();
-        //     ISavingsManager(savings).depositLiquidation(mAsset, minted);
+            // 5a. Send to SavingsManager
+            address savings = _savingsManager();
+            ISavingsManager(savings).depositLiquidation(mAsset, minted);
 
-        //     emit Liquidated(sellToken, mAsset, minted, bAsset);
-        // } else {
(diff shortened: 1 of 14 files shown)


package.json
```diff
@@ -37,7 +37,8 @@
         "test:file:fork": "yarn hardhat --config hardhat-fork.config.ts test",
         "test:file": "yarn hardhat test",
         "test:file:long": "LONG_TESTS=true yarn hardhat test",
-        "prepublishOnly": "yarn run compile"
+        "copy-types": "cp -R types/generated dist/types",
+        "prepublishOnly": "yarn run compile && yarn run copy-types"
```
alsco77, line 41: Super minor thing but you dont need the `run` with yarn

contracts/masset/liquidator/Liquidator.sol
```diff
@@ -119,9 +119,16 @@ contract Liquidator is ILiquidator, Initializable, ModuleKeysStorage, ImmutableM
...
+    function initialize() external initializer {
         IERC20(aaveToken).safeApprove(address(uniswapRouter), type(uint256).max);
         IERC20(compToken).safeApprove(address(uniswapRouter), type(uint256).max);
+    }
+
+    /**
+     * @notice Liquidator approves Uniswap to transfer Aave, COMP and ALCX tokens
+     * @dev to be called via the proxy proposeUpgrade function, not the constructor.
+     */
+    function upgrade() external {
```
alsco77, line 131: Not convinced of the need to store `alchemixToken` as an immutable, but also a minor point

alsco77 (approved):
You are doing a lot of good work with the tasks etc.

Only thing I am concerned about here is the lack of liquidity in the ALCX/eth pair (see screenshot) 
image

ALCX will be pulled weekly and is likely to be of a decent amount.. something to be aware of 

Potential solutions (if required):
 - Use the Sushiswap pair (same as Uniswap v2 interface) for ALCX->ETH
 - liquidate on a more frequent basis (every 1-3 days depending on value accruing)
 - Provide liquidity to the ALCX/ETH pair ourselves (20k or so)

contracts/masset/liquidator/Liquidator.sol
naddison36: It's needed because the Liquidator has a proxy. The proxy needs to give the approval to Uniswap to spend the ALCX tokens.
I can't move to the `initialize` function either as the existing proxy has already given the approval for AAVE and COMP to be spent by Umiswap. And as they are safeApproves, the will fail if run again.

package.json
naddison36: I'll make is consistent with the other scripts and remove the `run`s

PR #203, opened 2021-08-16
Liquid staking tests

alsco77:


contracts/governance/staking/GamifiedToken.sol
```diff
@@ -501,32 +464,33 @@ abstract contract GamifiedToken is
...
         (Balance memory oldBalance, uint256 oldScaledBalance) = _prepareOldBalance(_account);
-        require(oldBalance.raw >= _rawAmount, "ERC20: burn amount exceeds balance");
+        require(oldBalance.raw >= _rawAmount, "ERC20: burn amount > balance");
         unchecked {
             _balances[_account].raw = oldBalance.raw - SafeCast.toUint128(_rawAmount);
         }
 
         // 2. Change the cooldown percentage based on size of recent withdrawal
-        _balances[_account].cooldownMultiplier = SafeCast.toUint16(_cooldownPercentage / 1e16);
+        _balances[_account].cooldownMultiplier = SafeCast.toUint16(_cooldownPercentage / 1e18);
```
alsco77, line 477: Are you sure this is correct? If cooldown is 50% (5e17), we want to store it as "50", so dividing by 1e16 should be right

contracts/governance/staking/GamifiedToken.sol
```diff
@@ -501,32 +464,33 @@ abstract contract GamifiedToken is
...
         // e.g. stake 10 for 100 seconds, withdraw 5.
         //      secondsHeld = (100 - 0) * (10 - 1.25) = 875
         uint256 secondsHeld = (block.timestamp - oldBalance.weightedTimestamp) *
             (oldBalance.raw - (_rawAmount / 4));
         //      newWeightedTs = 875 / 100 = 87.5
-        uint256 newWeightedTs = secondsHeld / oldBalance.raw;
-        _balances[_account].weightedTimestamp = SafeCast.toUint32(block.timestamp - newWeightedTs);
+        uint256 newSecondsHeld = secondsHeld / oldBalance.raw;
+        uint32 newWeightedTs = SafeCast.toUint32(block.timestamp - newSecondsHeld);
+        _balances[_account].weightedTimestamp = newWeightedTs;
```
alsco77, line 487: Same logic but correct naming conventions, good call.

contracts/governance/staking/GamifiedToken.sol
```diff
@@ -1,13 +1,14 @@
...
 import { Initializable } from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
-import { HeadlessStakingRewards } from "../../rewards/staking/HeadlessStakingRewards.sol";
 import { ContextUpgradeable } from "@openzeppelin/contracts-upgradeable/utils/ContextUpgradeable.sol";
-import { SignatureVerifier } from "./deps/SignatureVerifier.sol";
 import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import { ILockedERC20 } from "./interfaces/ILockedERC20.sol";
+import { SignatureVerifier } from "./deps/SignatureVerifier.sol";
+import { HeadlessStakingRewards } from "../../rewards/staking/HeadlessStakingRewards.sol";
 import "./GamifiedTokenStructs.sol";
+import "./GamifiedManager.sol";
```
alsco77, line 11: Can we add a named import here

contracts/governance/staking/GamifiedTokenStructs.sol
```diff
@@ -38,3 +38,22 @@ struct Quest {
     /// Expiry date in seconds for the quest
     uint32 expiry;
 }
+
+struct SafetyData {
```
alsco77, line 42: How come these need to be in here? I am not sure the name makes sense given this was originally for the structs from the parent

contracts/governance/staking/StakedToken.sol
```diff
@@ -121,25 +110,33 @@ contract StakedToken is IStakedToken, GamifiedVotingToken {
...
     modifier onlyBeforeRecollateralisation() {
+        _onlyBeforeRecollateralisation();
+        _;
+    }
+
+    function _onlyBeforeRecollateralisation() internal view {
         require(
             safetyData.collateralisationRatio == 1e18,
-            "Function can only be called while fully collateralised"
+            "only while fully collateralised"
```
alsco77, line 120: Should capitalise

contracts/governance/staking/StakedToken.sol
```diff
@@ -231,19 +228,22 @@ contract StakedToken is IStakedToken, GamifiedVotingToken {
...
-        );
+        CooldownData memory oldCooldown = stakersCooldowns[_msgSender()];
         //      If we have missed the unstake window, or the user has chosen to exit the cooldown,
         //      then reset the timestamp to 0
         bool exitCooldown = _exitCooldown ||
-            block.timestamp > (nextCooldown + COOLDOWN_SECONDS + UNSTAKE_WINDOW);
+            block.timestamp > (oldCooldown.timestamp + COOLDOWN_SECONDS + UNSTAKE_WINDOW);
         if (exitCooldown) {
             stakersCooldowns[_msgSender()] = CooldownData(0, 0);
+            _exitCooldownPeriod(_msgSender());
```
alsco77, line 238: This line is not actually necessary, this will be picked up in the `_mintRaw` fn (see point 2. in mintRaw in GamifiedToken)

contracts/governance/staking/StakedToken.sol
```diff
@@ -231,19 +228,22 @@ contract StakedToken is IStakedToken, GamifiedVotingToken {
...
+            emit CooldownExited(_msgSender());
```
alsco77, line 239: Event could stay though

contracts/governance/staking/StakedToken.sol
```diff
@@ -231,19 +228,22 @@ contract StakedToken is IStakedToken, GamifiedVotingToken {
...
         } else {
-            stakersCooldowns[_msgSender()].timestamp = SafeCast.toUint128(nextCooldown);
+            //  Set new percentage so amount being cooled is the same as before the this stake.
+            //  new percentage = old percentage * old staked balance / (old staked balance + new staked amount)
+            uint256 stakedAmountOld = uint256(_balances[_msgSender()].raw);
+            uint256 newPercentage = (oldCooldown.percentage * stakedAmountOld) / (stakedAmountOld + _amount);
+            stakersCooldowns[_msgSender()].percentage = SafeCast.toUint128(newPercentage);
+            _enterCooldownPeriod(_msgSender(), newPercentage);
```
alsco77, line 246: Instead of entering a cooldownPeriod here, we should just pass the percentage into `mintRaw`. It will get handled there and save a call

alsco77 (commented):
See TG

contracts/governance/staking/StakedToken.sol
naddison36: If I remove this the `should end partial cooldown via staking` test fails. The `previousBalance` in the `DelegateVotesChanged` event goes from the correct 7,000 to 1,400 amount. This test initially stakes 7,000, enters a 80% cooldown and then stakes another 3,000 with the exit cooldown flag set to true.

contracts/governance/staking/GamifiedToken.sol
naddison36: I've written a test for this and you are right. I'll revert

PR #223, opened 2021-09-06
Vault deployment

naddison36:
- [x] vault upgrade to used `updateReward` modifier that calls an internal `_updateReward` function
- [x] set BoostDirector in mUSD vault to the newly deployed `0xBa05FD2f20AE15B0D3f20DDc6870FeCa6ACd3592` contract


alsco77 (approved):
LGTM

PR #242, opened 2021-09-23
Post Staking V2 tasks

naddison36:
- [x] Hardhat tasks to impersonate staking V1 txs
- [x] Hardhat tasks to interact and impersonate with staking V2 contracts
- [x] support for private transactions using the Taichi network
- [x] AAVE and COMP liquidation Hardhat task now using private transactions
- [x] getSigner can use a wallet from `PRIVATE_KEY` env var for testing
- [x] Added `exit` to `IIncentivisedVotingLockup interface`
- [x] Hardhat tasks `liq-trig` and `liq-trig-aave` tasks can be run as a fork test
- [x] Hardhat `rewards` task now catches error if it can't get ALCX price from Uniswap V3


tasks/ops.ts
```diff
@@ -154,6 +131,64 @@ task("quest-add", "Adds a quest to the staked token")
...
+            process.exit(0)
+        }
+        // filter users to just the migration quest
+        const migrationQuestId = 0
+        const completedMigrationQuests = queue.filter((quest) => quest.ethereumId === migrationQuestId)
+        const completedMigrationUsers = completedMigrationQuests.map((quest) => quest.userId)
+
+        // Need to filter out any users that completed the quest themselves
+        const hasCompletedPromises = completedMigrationUsers.map((user) => questManager.hasCompleted(user, migrationQuestId))
+        const hasCompleted = await Promise.all(hasCompletedPromises)
```
alsco77, line 168: Should add another line here for actually pinging the API to check for completion, and filter out the users who do not meet the criteria (as a secondary validation check)

tasks/stakingV2.ts
```diff
@@ -0,0 +1,213 @@
+import { subtask, task, types } from "hardhat/config"
+import { StakedTokenBPT__factory, StakedTokenMTA__factory, StakedToken__factory } from "types/generated"
+import { simpleToExactAmount } from "@utils/math"
+import { formatUnits } from "@ethersproject/units"
+import { getSigner } from "./utils/signerFactory"
+import { logTxDetails } from "./utils/deploy-utils"
+import { getChain, resolveAddress } from "./utils/networkAddressFactory"
+
+subtask("staked-stake", "Stake MTA or mBPT in V2 Staking Token")
```
alsco77, line 9: Subtasks can be called form other tasks, but not from test files, right?

tasks/stakingV2.ts
```diff
@@ -0,0 +1,213 @@
...
+            tx = await stakingToken["claimReward()"]()
+        }
+        const receipt = await logTxDetails(tx, `Claim earned MTA rewards`)
+        console.log(`Claimed ${formatUnits(receipt.events[0].args[2])} MTA rewards`)
+    })
+task("staked-claim").setAction(async (_, __, runSuper) => {
+    await runSuper()
+})
+
+subtask("staked-compound", "Stake any earned MTA rewards")
```
alsco77, line 129: No need to have a task for this, since it is not an administrative fn?

alsco77: Oh i see, for fork usage?

tasks/utils/taichi.ts
```diff
@@ -0,0 +1,87 @@
...
+        console.log(`nonce ${txUnsigned.nonce}`)
+        console.log(`gas limit ${txUnsigned.gasLimit}`)
+        console.log(`gas price ${txUnsigned.gasPrice}`)
+
+        const txRaw = await signer.signTransaction(txUnsigned)
+
+        nonce += 1
+    }
+
+    return sendBundledRawTransactions(rawTxs, BN.from(block.number))
```
alsco77, line 86: Nice

alsco77 (changes requested):
Will wait for that validation change before merging in. Otherwise, looks great and comprehensive task impl 👍

tasks/ops.ts
naddison36: Good idea. I'll add

tasks/stakingV2.ts
naddison36: Yes, subtasks can be called from tasks. eg https://github.com/mstable/mStable-contracts/blob/b4fbb024b617344d0d1671f31d76a8025320d360/tasks/weekly.ts#L38 

I think if we want a test to run something in a task we move the logic into a function. We've done that a number of times already

tasks/stakingV2.ts
naddison36: Yes, used for impersonating users for Support

tasks/utils/taichi.ts
naddison36: This will be better once Defender Relay supports `signTransaction`. At the moment we have to use a local wallet to sign transactions.

tasks/ops.ts
naddison36: I've added a check against Quest Book's `user` query for each user that is claiming quest completion.

alsco77 (approved):

PR #244, opened 2021-10-01
Fixes to mUSD and mBTC snaps

naddison36:


tasks/mBTC.ts
```diff
@@ -60,9 +60,7 @@ task("mBTC-snap", "Get the latest data from the mBTC contracts")
         const { ethers, network } = hre
 
         let exposedValidator
-        if (network.name !== "mainnet") {
```
rafaelugolini, line 63: Should we improve `getChain` to include `hardhat` and avoid using the `network.name` to be consistent?

rafaelugolini (approved):

rafaelugolini (approved):

PR #262, opened 2021-10-27
feat: add NonPeggedFeederPool deploy task

dimsome:


tasks/utils/tokens.ts
```diff
@@ -209,6 +210,31 @@ export const BUSD: Token = {
     vault: "0xD124B55f70D374F58455c8AEdf308E52Cf2A6207",
 }
 
+// NonPeggedFeederPool contains priceGetter
+export const RAI: Token = {
+    symbol: "RAI",
+    address: "0x03ab458634910aad20ef5f1c8ee96f1d6ac54919",
+    chain: Chain.mainnet,
+    platform: Platform.Aave,
+    integrator: "",
```
alsco77, line 219: Will need to deploy an integrator and link the address here in order for it to be deposited onto AAVE.

alsco77 (approved):
Aside from the comment, and assuming `0x07210B8871073228626AB79c296d9b22238f63cE` is the official price getter, LGTM

PR #264, opened 2021-10-28
feat: Packed preferences

alsco77:
Changes:
 - Changes scale to 200 to support uint8
 - Packs 16 preferences to an arr
 - Set up dummy dialId of 0
 - Loops through user prefs while dialId > 0
 - Enforces all dialIds to be > 0
 - Runs linter via husky
 - Fix failing test cases

Design choice:

Assumption: dialId should correspond to position in array, and not be used as offset.
When packing the preferences to single arr, it makes sense to loop through until some breaking point (either 16 entries, or first dial being 0). IN order to do this, I introduced a dud dial at position 0, and then we loop through the preferences while id != 0.
We could however, when setting preferences, signal the final entry by setting dialId to be 256 on the end of the array. This would mean we don't need to worry about the empty entry (which it emits during an event).

Open question:
 - Use 256 or 0 as breaker
 - If using 0, do we emit this as part of the rewardsDistributed event?

### DIFF
--- a/.github/workflows/test-emissions.yaml
+++ b/.github/workflows/test-emissions.yaml
@@ -21,4 +21,4 @@ jobs:
                       ${{ runner.os }}-yarn-
             - run: yarn
             - run: yarn compile
-            - run: yarn test:file ./test/emissions/**/*.spec.ts ./test/emissions/*.spec.ts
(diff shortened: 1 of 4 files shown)


naddison36 (approved):
Looks good. Thanks

PR #268, opened 2021-11-04
Fixed 10% to staking contracts

naddison36:


contracts/emissions/ForwarderRecipient.sol
```diff
@@ -0,0 +1,95 @@
...
+import { IERC20, SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title  ForwarderRecipient
+ * @author mStable
+ * @notice transfers any received reward tokens to another contract or account.
+ * @dev     VERSION: 1.0
+ *          DATE:    2021-10-28
+ */
+contract ForwarderRecipient is
```
alsco77, line 20: My vote is that this `ForwarderRecipient` is:
a) Immutable (i.e. not a proxy)
b) `Ownable` by whoever is in charge of the recipient - allowing them to call `setEndRecipient`
c) `onlyGovernor` can still be in charge of rewardsDistributor

PR #284, opened 2021-11-23
chore: adds commitizen config to the project

doncesarts:
Adds default configuration of commitizen in order to use conventional commit
Example to run 
`npm run commit` 
A prompt will appear in order to enforce conventional commits

Screenshot 2021-11-23 at 18 06 50
s


doncesarts:
There is  husky already in place so no need at the moment for this

PR #288, opened 2021-11-25
test: adds tests to EmissionsController.topLineEmission function (9702)

doncesarts:
- Change consturctor, as now the it is required to  numbers of 9 digits on the configuration to avoid loosing precision.
    From `A = _config.A * 1e12;`  to `A = _config.A * 1e3;`
- Fix overflow execption while calculating epochDelta 
```
// Uncaught Error: VM Exception while processing transaction: reverted with panic code 0x11 (Arithmetic operation underflowed or overflowed outside of an unchecked block)
 uint128 epochDelta = (epoch - epochs.startEpoch) * 1e18; 
```

### DIFF
--- a/contracts/emissions/EmissionsController.sol
+++ b/contracts/emissions/EmissionsController.sol
@@ -110,7 +110,7 @@ contract EmissionsController is IGovernanceHook, Initializable, ImmutableModule 
     mapping(address => VoterPreferences) public voterPreferences;
 
     event AddedDial(uint256 indexed id, address indexed recipient);
-    event UpdatedDial(uint256 indexed id, bool diabled);
+    event UpdatedDial(uint256 indexed id, bool disabled);
     event AddStakingContract(address indexed stakingContract);
 
     event PeriodRewards(uint256[] amounts);
@@ -160,7 +160,7 @@ contract EmissionsController is IGovernanceHook, Initializable, ImmutableModule 
         address[] memory _stakingContracts
     ) external initializer {
         uint256 len = _recipients.length;
-        require(_notifies.length == len && _caps.length == len, "Initialize args mistmatch");
+        require(_notifies.length == len && _caps.length == len, "Initialize args mismatch");
 
         // 1.0 - Add each of the dials
         for (uint256 i = 0; i < len; i++) {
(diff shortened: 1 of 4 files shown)


contracts/emissions/EmissionsController.sol
```diff
@@ -203,19 +203,19 @@ contract EmissionsController is IGovernanceHook, Initializable, ImmutableModule
      * @return emissionForEpoch Units of MTA to be distributed at this epoch
      */
     function topLineEmission(uint32 epoch) public view returns (uint256 emissionForEpoch) {
-        // e.g. week 1, A = -166000e12, B = 180000e12, C = -180000e12, D = 166000e12
-        // e.g. epochDelta = 1e18
-        uint128 epochDelta = (epoch - epochs.startEpoch) * 1e18;
-        // e.g. x = 1e18 / 312e6 = 3205128205
-        int256 x = SafeCast.toInt256(epochDelta / EPOCHS);
+        require(epochs.startEpoch < epoch && epoch<= epochs.startEpoch + 312 , "Wrong epoch number");
```
doncesarts, line 206: Not sure about the upper limit, epoch<= epochs.startEpoch + 312 , it avoids to return negative values with the current configuration,  but this could change if the configuration changes

test/emissions/emission-controller.spec.ts
```diff
@@ -20,28 +20,53 @@ import {
 import { currentWeekEpoch, increaseTime, getTimestamp, increaseTimeTo, startWeek } from "@utils/time"
 
 const defaultConfig = {
-    A: -166000,
-    B: 180000,
-    C: -180000,
-    D: 166000,
+    A: -166000000000000,
```
doncesarts, line 23: @dimsome  are you ok with the values updated ?

contracts/emissions/EmissionsController.sol
naddison36: That looks correct to me

test/emissions/emission-controller.spec.ts
naddison36: This matches the spreadsheet so looks good

naddison36 (approved):
LGTM. Thanks

PR #295, opened 2021-12-01
Changed getEpochVotes to getDialVotes

naddison36:
* Changed `getEpochVotes` to `getDialVotes` which just gets the latest weighted votes for each dial.
  This was required as `getEpochVotes` assumed there was a `voteHistory` record for every week. This was not the case if a dial was disabled.
* Added dial for the FEI Vault


doncesarts (approved):
The  change looks good to me

alsco77 (approved):

PR #305, opened 2021-12-08
Dud integration

dimsome:
- [x] Write fork test for 1 Feeder Pool
- [x] Write fork test to migrate second Feeder Pool

### DIFF
--- a/contracts/masset/peripheral/DudIntegration.sol
+++ b/contracts/masset/peripheral/DudIntegration.sol
@@ -20,6 +20,13 @@ import { AbstractIntegration } from "./AbstractIntegration.sol";
 contract DudIntegration is AbstractIntegration {
     using SafeERC20 for IERC20;
 
+    // TODO - support virtual amount here
+    uint256 pseudoDeposited = 0;
+
+    // increase on deposit
+    // decrease on totalWithdraw
+    // return balance from checkbalance
+
     /**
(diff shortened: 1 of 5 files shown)


dimsome:
PR contains 
- `dudIntegration.sol` is the replacement contract for the Integration Contract that a Feeder Pool uses. (to go from Integration to "no"-integration
- `dudPlatform.sol` to temporally act as the platform, needed because otherwise, the migration will fail. Can be later called by governance to `clear` which short-circuits all following deposits and sends all back to the `dudIntegration.sol`, soft of as a cache
- units tests
- fork tests, migration of the second feeder fails (added skip to test): this is because some mUSD is borrowed and cannot be withdrawn, the function `migrateBasset` can only fully withdraw. Additionally, minting of the cymUSD fails with `minting is pause`.

My recommendation: 
- Migrate one feeder pool -> that will increase borrow apy -> might incentives to repay the loan
- in the meantime, investigate the minting issue. If resolved, can add liquidity from somewhere (~amount that is borrowed) and migrate the second feeder pool. This will squeeze the borrower and either force a repayment or have some good yield.

dimsome:
moved variable `cleared` and `clear()` function to `DudIntegration.sol` for fewer contract calls and less gas usage.

naddison36 (approved):
I've made a few small changes.
Looks good to go.

PR #340, opened 2022-02-22
SaveWrapper standardisation

naddison36:
* `SaveWrapper` approve functions changed to use mStable `Governor` (Protocol DAO) or `Keeper` (Relay) instead of an `owner`.
* SaveWrapper tasks updated to use token symbols rather than addresses.


tasks/SaveWrapper.ts
```diff
@@ -1,80 +1,116 @@
...
+    .addParam(
+        "spender",
+        "Token symbol of the mAsset or address type. eg mUSD, mBTC, feederPool, savings or vault",
+        undefined,
+        types.string,
+        false,
+    )
+    .addOptionalParam("speed", "Defender Relayer speed param: 'safeLow' | 'average' | 'fast' | 'fastest'", "fast", types.string)
+    .setAction(async (taskArgs, hre) => {
+        if (!taskArgs.spender) {
```
doncesarts, line 98: it is good to add this kind of checks for future references , as it is easy to forget what should be passed or not

doncesarts (approved):
It is a Good improvement

PR #342, opened 2022-03-01
Feature share revenue

dimsome:
Forked `RevenueBuyBack` contract to `RevenueSplitBuyBack` that sends a portion of the mAssets revenue to treasury and the rest through a buy-back of reward tokens. The buy-back does a mAsset redeem to a bAsset and then a Uniswap V3 swap to the rewards token, eg MTA.

The buy-back moves the min amounts to parameters of the `buyBackRewards` function instead of being stored in contract storage. This is to better protect against sandwich attacks. Only the keeper or governor can now execute the `buyBackRewards` function as setting the min amounts needs to come from a trusted account.

Change details

* `buyBackRewards` has an array of `minBassetsAmounts`, `minRewardsAmounts` and `uniswapPaths` added as arguments to the existing array of `mAssets` argument.
* The portion of the revenue sent to the treasury account is set with `setTreasuryFee`.
* The treasury account is set with `setTreasury`
* `setMassetConfig` has been changed to just `mapBasset` as there's now just a mapping from mAssets to bAssets. The min bAsset and rewards prices along with the uniswap path are no longer stored in contract storage.

TODO
* fork tests

### DIFF
--- a/contracts/buy-and-make/RevenueSplitBuyBack.sol
+++ b/contracts/buy-and-make/RevenueSplitBuyBack.sol
@@ -14,7 +14,7 @@ import { IUniswapV3SwapRouter } from "../peripheral/Uniswap/IUniswapV3SwapRouter
 /**
  * @title   RevenueSplitBuyBack
  * @author  mStable
- * @notice  Uses protocol revenue to buy MTA rewards for stakers. Updated Version sends some protocol fees to treasury.
+ * @notice  Uses governance fees to buy MTA rewards for stakers. Updated Version sends some governance fees to treasury.
  * @dev     VERSION: 2.0
  *          DATE:    2022-04-17
  */
@@ -31,7 +31,7 @@ contract RevenueSplitBuyBack is IRevenueRecipient, Initializable, ImmutableModul
     event DonatedRewards(uint256 totalRewards);
     event MappedBasset(address indexed mAsset, address indexed bAsset);
     event AddedStakingContract(uint16 stakingDialId);
-    event ProtocolFeeChanged(uint256 protocolFee);
+    event TreasuryFeeChanged(uint256 treasuryFee);
     event TreasuryChanged(address treasury);
 
     /// @notice scale of the `minMasset2BassetPrice` and `minBasset2RewardsPrice` configuration properties.
@@ -49,8 +49,8 @@ contract RevenueSplitBuyBack is IRevenueRecipient, Initializable, ImmutableModul
     /// @notice Emissions Controller dial ids for all staking contracts that will receive reward tokens.
     uint256[] public stakingDialIds;
 
-    /// @notice ProtocolFee, how much does go back to the Treasury? 100% = 1e18
-    uint256 public protocolFee;
+    /// @notice percentage of governance fees that is sent to the Treasury where 100% = 1e18
+    uint256 public treasuryFee;
 
     /// @notice address the Treasury fees are transferred to.
     address public treasury;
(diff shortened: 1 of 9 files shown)


contracts/buy-and-make/RevenueSplitBuyBack.sol
```diff
@@ -0,0 +1,350 @@
...
+                    uniswapPaths[i],
+                    address(this),
+                    block.timestamp,
+                    bAssetAmount,
+                    minRewardsAmounts[i]
+                );
+                rewardsAmount = UNISWAP_ROUTER.exactInput(param);
+            }
+
+            emit BuyBackRewards(mAssets[i], mAssetToTreasury, mAssetsSellAmount, rewardsAmount);
```
doncesarts, line 212: mAssets[i] is called more than once, is it worthy to save it as a variable. ? 
mAsset = mAssets[i]

contracts/buy-and-make/RevenueSplitBuyBack.sol
```diff
@@ -0,0 +1,350 @@
...
+        bassets[_mAsset] = _bAsset;
+
+        emit MappedBasset(_mAsset, _bAsset);
+    }
+
+    /**
+     * @notice Sets the percentage of governence fees to be sent to treasury.
+     * @param _treasuryFee percentage of governence fees to be sent to treasury where 100% = 1e18.
+     */
+    function setTreasuryFee(uint256 _treasuryFee) external onlyGovernor {
```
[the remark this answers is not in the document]
doncesarts: I just saw the the internal function ..

test-fork/buy-and-make/revenue-split.spec.ts
```diff
@@ -0,0 +1,152 @@
...
+    SavingsManager,
+    SavingsManager__factory,
+} from "types/generated"
+import { Account } from "types/common"
+import { encodeUniswapPath } from "@utils/peripheral/uniswap"
+import { keccak256 } from "@ethersproject/keccak256"
+import { toUtf8Bytes } from "ethers/lib/utils"
+
+const keeperKey = keccak256(toUtf8Bytes("Keeper"))
+console.log(`Keeper ${keeperKey}`)
```
doncesarts, line 26: remove console log

test/buy-and-make/revenue-split-buy-back.spec.ts
```diff
@@ -0,0 +1,640 @@
...
+        await revenueBuyBack.connect(sa.governor.signer).mapBasset(mUSD.address, bAsset1.address)
+        await revenueBuyBack.connect(sa.governor.signer).mapBasset(mBTC.address, bAsset2.address)
+    }
+
+    before(async () => {
+        const accounts = await ethers.getSigners()
+        mAssetMachine = await new MassetMachine().initAccounts(accounts)
+        sa = mAssetMachine.sa
+
+        await setupRevenueBuyBack(treasuryFee)
```
doncesarts, line 131: is it necessary to add  await setupRevenueBuyBack(treasuryFee) before all tests? 
I think that it is initialised anyway  within each  internal`describe`

doncesarts (commented):
I raised some questions , in overall it looks good

contracts/buy-and-make/RevenueSplitBuyBack.sol
naddison36: mAssets is calldata so it's not an expensive storage variable. Having said that, saving gas is always good so I tried adding `mAsset = mAssets[i];` to see what the difference was. Unfortunately, this resulted in the stack too deep error.

```
CompilerError: Stack too deep, try removing local variables.
   --> contracts/buy-and-make/RevenueSplitBuyBack.sol:243:21:
    |
243 |                     minBassetsAmounts[i],
    |                     ^^^^^^^^^^^^^^^^^
```

I could move part of the logic in the loop out into an internal function to avoid this error but I think we'll loose any gas savings if we do this.

`buyBackRewards` is only called once a week so probably not spending hours saving a small amount of gas.

contracts/buy-and-make/RevenueSplitBuyBack.sol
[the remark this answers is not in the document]
naddison36: I originally had a check for a small fee but I took it out as I think a valid use case is setting the treasury fee to zero. In this scenario, all protocol fees go to the stakers like they currently do.

test-fork/buy-and-make/revenue-split.spec.ts
naddison36: Good pick up. Done

naddison36 (approved):

PR #347, opened 2022-04-06
Feature/ierc4626 vault

doncesarts:
Introduces new spec 4626 to Saving Contracts. 
- contracts/interfaces/IERC4626Vault.sol

Creates version 4 of ISavingsContract , with two new non 4626 methods , deposit and mint with a referrer . 

Updates  use of ISavingsContractV3. to  ISavingsContractV4 across the code base.  

Unit test for spec 4626 to Saving Contracts. 

Fork tests for spec 4626 to Saving Contracts. 

Deployment tasks for spec 4626 to Saving Contracts.


contracts/savings/SavingsContract.sol
```diff
@@ -743,4 +725,301 @@ contract SavingsContract is ISavingsContractV3, Initializable, InitializableToke
...
+     * @param receiver The account the underlying assets will be transferred to.
+     * @param owner The account that owns the vault shares to be burnt.
+     * @return assets The amount of underlying assets that were transferred to the receiver.
+     */
+    function redeem(
+        uint256 shares,
+        address receiver,
+        address owner
+    ) external override returns (uint256 assets) {
+        require(shares > 0, "Must withdraw something");
```
bitcoinbrisbane, line 947:
```suggestion
        require(shares > 0, "Must redeem something");
```

naddison36: This is the same error message as the old `redeemCredits` function so we'll just leave it with `Must withdraw something` in this ERC4626 implementation.

doncesarts:
4626 here we go

PR #350, opened 2022-04-13
Deposit mBPT in Staked Token BPT (stkBPT) in Balancer's new mBPT Gauge (mBPT-Gauge)

naddison36:
## Functional Changes

The functional changes are pretty simple. Instead of the [mStable MTA/WETH Staking BPT (mBPT)](https://etherscan.io/address/0xe2469f47ab58cf9cf59f9822e3c5de4950a41c49) being held in the [Staking Token BPT (stkBPT)](https://etherscan.io/address/0xeFbe22085D9f29863Cfb77EEd16d3cC0D927b011) contract, it would be deposited into the new [Balancer mBPT Gauge Deposit (mBPT-gauge)](https://etherscan.io/address/0xbec2d02008dc64a6ad519471048cf3d3af5ca0c5). 

Detailed functional changes to the `StakedTokenBPT` contract:
- Add immutable `balancerGauge` variable that contains the address of the new [Balancer mBPT Gauge Deposit (mBPT-gauge)](https://etherscan.io/address/0xbec2d02008dc64a6ad519471048cf3d3af5ca0c5) and implements the new IBalancerGauge interface.
- As `balancerGauge` is immutable, it must be set in the constructor.
- A new function added to
	- approve the new Balancer gauge contract to transfer mBPT from the `StakedTokenBPT` contract.
	- Migrate all the mBPT held in the `StakedTokenBPT` contract to the new Balancer gauge.
- Modify the `setBalRecipient` function so the default rewards receiver is set on the new gauge instead of the old `balRecipient` storage variable.
- Modify the `convertFees` function so 
	- the mBPT balance of the `StakedTokenBPT` contract is retrieved from the new gauge rather than the `StakedTokenBPT`'s mBPT balance.
	- pending mBPT fees are withdrawn from the gauge before the balancer pool is used to swap mBPT for MTA.
- Modify the internal `_transferAndStake` function so mBPT is transferred to the new gauge after mBPT is transferred into the `StakedTokenBPT` contract.
- Modify the internal `_withdrawStakedTokens` function so mBPT is withdraw from the gauge before mBPT is transferred to the withdrawal recipient.
- Add new internal `_balanceOfStakedTokens` function that gets the mBPT balance staked in `StakedTokenBPT` from the new Balancer gauge.

Functional changes to the parent `StakeToken` contract:
- Make internal function `_transferAndStake` virtual so it can be overridden by `StakedTokenBPT`.
- Add new internal `_withdrawStakedTokens` function that transfer's mBPT to the recipient by default but can be overridden by `StakedTokenBPT` to withdraw mBPT from the new Balancer gauge.
- Add new internal `_balanceOfStakedTokens` function that gets the balance of the staked MTA or mBPT tokens. By default it gets the balance of the staked tokens held in the staking contract but can be overridden by `StakedTokenBPT` which gets the balance from the new Balancer gauge.

New `IBalancerGauge` interface with the following functions
- deposit
- withdraw
- set_rewards_receiver

## Non-functional changes

The `StakedTokenBPT` contract was already at the limit of allowed contract size so adding the extra functional changes was always going to push it over.

The following changes were done to get the contract back under size
- `HeadlessStakingRewards` changed to use mStable's `Context` contract based off Open Zeppelin 2.5 rather than Open Zeppelin's 4.3.3 `ContextUpgradeable` contract. This drops the internal `__Context_init` and `__Context_init_unchained` functions that don't do anything except use the `initializer` modifier. The `initializer` modifier no longer needed as the `initialize` function can be rerun. This change significantly reduces the contract size.
- Removed inheritance of `ContextUpgradeable` from `GamifiedToken` as `Context` is already inherited in `HeadlessStakingRewards`.
- Removed inheritance of `Initializable` from `GamifiedToken` as its internal `__GamifiedToken_init` function is not using the `initializer` modifier.
- Remove the internal `__GamifiedVotingToken_init` function from `GamifiedVotingToken` as it doesn't do anything.
- Remove inheritance of `Initializable` from `GamifiedVotingToken` as the `initializer` modifier on `__GamifiedVotingToken_init` is not needed.
- Modified `StakedToken` so the `UNSTAKE_WINDOW` variable is a constant rather than an immutable which allows the `_unstakeWindow` param to be removed from the constructor.
- Removed the `initializer` modified from the internal `StakedToken.__StakedToken_init` function.
- Reduced the revert error messages sizes in `StakedToken`
	- `Only Recollateralisation Module` to `Only Recollateralisation`
	- `Only while fully collateralised` to `Only while collateralised`
	- `Not a whitelisted contract` to `Not whitelisted`
- Reduced the revert error message sizes in `StakedTokenBPT`
	- `Must have something to `convert` to `no fees`
	-` MTA in wrong place` to `not MTA`
	- `Must sell correct amount of BPT` to `< min BPT`
	- `Must receive tokens` to `< min MTA`
	-` more than a million units` to `> mil`
	- `Max 1 update per 14 days` to `< 14 days`
- Modified `StakedTokenBPT`
	- Removed `claimBal` function and `BalClaimed` event as it can be claimed directly from the balancer gauge
	- Removed `governorOrKeeper` modifier and `setKeeper` function as there is now a `onlyKeeperOrGovernor` modifier inherited from the `ImmutableModule` contract.
	- Removed `_unstakeWindow` param from the constructor
	- `StakedTokenBPT` storage variables `balRecipient` and `keeper` are made private as they are no longer used. They removes two view functions from the ABI. They aren't deleted as the slots need to be preserved.
	- Ideally a new migrate function would be added but to reduce the contract size the existing `initialize` function is repurposed. The original `initialize` functionality is kept so there isn't too many changes to the unit tests. If `rewardsDistributor` has already been set like on mainnet then skip doing the contract initialisation. Any mBPT held by the `StakedTokenBPT` contract is deposited into the new Balancer Gauge.

## Storage

The non-functional changes impacted the storage slots. As `StakedTokenBPT` is a proxied contract who's storage is in the proxy, it's critical that the new contract's storage slots align with the old one.

- Using mStable's `Context` contract based off Open Zeppelin 2.5 rather than Open Zeppelin's 4.3.3 `ContextUpgradeable` contract changes the storage slots. The first slot is used for the private `_initialized` and `_initializing` storage variables in the parent `Initializable` contract.
- `ContextUpgradeable` also used 50 slots with a private `__gap` storage variable.
- In order to preserve these slots so the following slots align, a `SlotFiller51` contract was added which fills 51 slots with a private `__gap` storage variable. This is added to the inheritance of the `GamifiedToken` contract.
- `StakedTokenBPT` storage variables `balRecipient` and keeper are made private so there is no view functions in the ABI, which reduces the contract size. They can't be deleted as their slot positions need to be preserved so the following storage variables align with their old storage slots.

The storage slots before `StakedTokenBPT` is upgraded
StakedTokenBPTv1

The storage slots after `StakedTokenBPT` is upgraded
StakedTokenBPTv2


contracts/governance/staking/StakedToken.sol
```diff
@@ -301,15 +306,24 @@ contract StakedToken is GamifiedVotingToken, InitializableReentrancyGuard {
...
             emit Withdraw(_msgSender(), _recipient, _amount);
         }
     }
 
+    /**
+     * @dev Transfers an `amount` of staked tokens to the withdraw `recipient`. eg MTA or mBPT.
+     * Can be overridden if the tokens are held elsewhere. eg in the Balancer Pool Gauge.
+     */
+    function _withdrawStakedTokens(address _recipient, uint256 amount) internal virtual {
+        STAKED_TOKEN.safeTransfer(_recipient, amount);
```
doncesarts, line 324: Should the ` emit Withdraw(_msgSender(), _recipient, _amount);` be included in this method ?
Second thought , if the method is override and the event is not emitted then the behaviour would not be constant.

contracts/governance/staking/StakedTokenBPT.sol
```diff
@@ -51,81 +55,75 @@ contract StakedTokenBPT is StakedToken {
...
         bytes32 _symbolArg,
         address _rewardsDistributorArg,
-        address _balRecipient,
         uint256 _priceCoefficient
-    ) external initializer {
-        __StakedToken_init(_nameArg, _symbolArg, _rewardsDistributorArg);
-        balRecipient = _balRecipient;
-        priceCoefficient = _priceCoefficient;
-    }
+    ) external {
```
doncesarts, line 102: was the removal of  modifier `initializer` intentional ?

doncesarts: I think that i get it,  the init method can be called again, but the reward distributor can only be set once.

contracts/governance/staking/StakedTokenBPT.sol
```diff
@@ -250,7 +243,41 @@ contract StakedTokenBPT is StakedToken {
...
+    /***************************************
+                BALANCER POOL GAUGE
+    ****************************************/
+
+    /**
+     * @dev Transfers staked tokens from sender to this staking token contract,
+     * deposits staked tokens in the Balancer Pool Gauge to earn rewards and
+     * finally call `_settleStake`.
+     */
+    function _transferAndStake(
```
doncesarts, line 259: Should be there any validation to amount or delegatee ?  is there any limit that should be consider ?

contracts/governance/staking/StakedToken.sol
naddison36: It could be moved into `_withdrawStakedTokens` but there's a risk the implementation will not emit the event so I left the emits in `_withdraw`

contracts/governance/staking/StakedTokenBPT.sol
naddison36: Ideally, I would leave the `initialize` function with the `initializer` modifier but in order to get the contract size down I had to repurpose the `initialize` function to also do the migration.

contracts/governance/staking/StakedTokenBPT.sol
```diff
@@ -165,49 +168,39 @@ contract StakedTokenBPT is StakedToken {
...
+     * @dev Called by `StakedToken._withdraw` to add early withdrawal fee charged in the staking token mBPT.
+     * @param _fees Units of staking token mBPT.
      */
-    function _notifyAdditionalReward(uint256 _additionalReward) internal override {
-        require(_additionalReward < 1e24, "more than a million units");
+    function _notifyAdditionalReward(uint256 _fees) internal override {
+        require(_fees < 1e24, "> mil");
 
-        pendingBPTFees += _additionalReward;
+        pendingBPTFees += _fees;
```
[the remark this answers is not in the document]
naddison36: This code is very confusing. In the `StakedTokenBPT` contract the `_notifyAdditionalReward` function takes an amount of mBPT tokens while the `_notifyAdditionalReward` function in the inherited contract `HeadlessStakingRewards` takes an amount of MTA. I think the param name `_additionalReward` is misleading as it's not an MTA rewards amount that is passed into `_notifyAdditionalReward`.

contracts/governance/staking/StakedTokenBPT.sol
naddison36: Good question. The old code didn't do any validation on the amount staked. The user can stake as much as they want. If they try and stake more than they have then the transfer from the msg sender into the staking contract will fail.

There is an existing check at the start of `StakedToken._settleStake` for a zero amount staked.

There is also no restriction on who the stkBPT tokens are delegated to. We could add basic checks like no ZERO addresses but what about other address like address(1)? Given we are already at the contract size limit, I'd prefer not to validate the `_delegatee` is a zero address.

PR #354, opened 2022-04-21
maxWithdraw and maxRedeem param name changes

naddison36:
- `maxWithdraw` and `maxRedeem` param was changed from `caller` to `owner` in the final spec. I updated `imbtc-mainnet-22`, `imusd-mainnet-22`, `imusd-polygon-22` and `SavingsContract`. 
- Completed Natspec for some of the new ERC-4626 functions.


doncesarts (approved):
Amazing , thanks for the review of the natspec

PR #362, opened 2022-05-26
Small gas savings

naddison36:


.eslintrc.js
```diff
@@ -12,6 +12,8 @@ module.exports = {
         "import/resolver": {
             alias: {
                 map: [
+                    ["@task", "./task"],
```
doncesarts, line 15: Very  good, I just had an issue with the npm package that is generated, the links are broken.  But that is another issue

tushargargght (approved):
Good catch

naddison36:
This needs to be rebased before merge

PR #370, opened 2022-07-07
fix: remove disabled dials from output

dimsome:
added check to remove looping over disabled dials. Disabled dials will not produce a console output anymore.

### DIFF
--- a/tasks/dials.ts
+++ b/tasks/dials.ts
@@ -223,8 +223,10 @@ task("dials-snap", "Snaps Emissions Controller's dials")
 
         latestDialVotes.forEach(async (vote, dialId) => {
             const dialData = dialsData[dialId]
-            // 1.1- Get the weighted votes as a percentage of the total weighted votes across all dials
-            const voteWeight = percentToWeight(totalDialVotes.eq(0) ? BN.from(0) : vote.mul(10000).div(totalDialVotes))
+            // if the dial is disabled assign 0 to the vote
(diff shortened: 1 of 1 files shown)


tasks/dials.ts
```diff
@@ -115,9 +115,9 @@ const dialsDetailsToString = (dialsDetails: Array<DialDetails>) =>
     dialsDetails
         .map(
             (dd, i) =>
-                `${dialNames[i].padStart(21)}\t${usdFormatter(dd.voteWeight, 18, 5, 2)}\t${usdFormatter(dd.distributed)}\t${usdFormatter(
-                    dd.donated,
-                )}\t ${usdFormatter(dd.rewards)}`,
```
doncesarts, line 120: you can remove the index "i" , it is not used.
but question how it avoids the disabled ones ? i would expect a filter or similar

dimsome: right, forgot. The `i` was used before.

The filter happens in the calculation, disabled dials are skipped in the loop. Line 225

tasks/dials.ts
```diff
@@ -221,10 +221,11 @@ task("dials-snap", "Snaps Emissions Controller's dials")
```
doncesarts, line 221: it is just styling but i would rather see a filter latestDialVotes.filter().forEach
than a return inside the foreach

dimsome: Not really possible on that array without creating mismatched indexes, but reworked to add a filter where the string is built.

PR #377, opened 2022-11-30
fix claim MTA from StakedTokenBPT

naddison36:


doncesarts (approved):
I reviewed all references to _userPriceCoeff and _getPriceCoeff, this change looks good.

PR #380, opened 2023-03-09
feat: adds mta redeemer

doncesarts:
- Adds smart contract that allows to redeem MTA for WETH

### DIFF
--- a/contracts/shared/MetaTokenRedeemer.sol
+++ b/contracts/shared/MetaTokenRedeemer.sol
@@ -3,6 +3,7 @@ pragma solidity 0.8.6;
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { SafeCastExtended } from "../shared/SafeCastExtended.sol";
 
 /**
  * @notice  Allows to redeem MTA for WETH at a fixed rate.
@@ -12,14 +13,17 @@ import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.s
  */
 contract MetaTokenRedeemer {
     using SafeERC20 for IERC20;
+    struct RegisterPeriod {
+        uint32 start;
+        uint32 end;
+    }
 
     address public immutable MTA;
     address public immutable WETH;
     uint256 public immutable PERIOD_DURATION;
-    uint256 public periodStart;
-    uint256 public periodEnd;
-    uint256 public totalFunded;
-    uint256 public totalRegistered;
+    RegisterPeriod public registerPeriod;
+    uint128 public totalFunded;
+    uint128 public totalRegistered;
     mapping(address => uint256) public balances;
 
     /**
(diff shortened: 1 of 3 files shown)


contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,72 @@
...
+ * @notice  Allows to redeem MTA for WETH at a fixed rate.
+ * @author  mStable
+ * @dev     VERSION: 1.0
+ *          DATE:    2023-03-08
+ */
+contract MetaTokenRedeemer {
+    using SafeERC20 for IERC20;
+
+    uint256 public constant RATE_SCALE = 1e18;
+    address public immutable MTA;
```
naddison36, line 21: We will only ever deploy this on mainnet so it'd be easier to just have these as constants

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,72 @@
...
+    function fund(uint256 amount) external {
+        IERC20(WETH).safeTransferFrom(msg.sender, address(this), amount);
+        emit Funded(msg.sender, amount);
+    }
+
+    /// @notice Redeems MTA for WETH at a fixed rate.
+    /// @param fromAssetAmount a parameter just like in doxygen (must be followed by parameter name)
+    /// @return toAssetAmount The amount of WETH received.
+    function redeem(uint256 fromAssetAmount) external returns (uint256 toAssetAmount) {
+        IERC20(MTA).safeTransferFrom(msg.sender, address(this), fromAssetAmount);
```
naddison36: we could add an override so the MTA could be transferred from an approved account. I think it's fine to leave it like this for now.

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,72 @@
...
+        uint256 _rate
+    ) {
+        MTA = _mta;
+        WETH = _weth;
+        RATE = _rate;
+    }
+
+    /// @notice Funds the contract with WETH.
+    /// @param amount Amount of WETH to be transfer to the contract
+    function fund(uint256 amount) external {
```
naddison36, line 64: I was initially thinking we should restrict who can fund but since the payout is based on the rate it doesn't impact how much an MTA holder gets.
It creates some flexibility that Treasury assets could be contributed in multiple transfers.
If more ETH is funded than the circulating supply of MTA * rate, then that ETH will be forever stuck.

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,103 @@
...
+ * @dev     VERSION: 1.0
+ *          DATE:    2023-03-08
+ */
+contract MetaTokenRedeemer {
+    using SafeERC20 for IERC20;
+
+    address public immutable MTA;
+    address public immutable WETH;
+    uint256 public immutable PERIOD_DURATION;
+    uint256 public periodStart;
```
naddison36: timestamps like `periodStart` and `periodEnd` will fit in uint32 so can take just one slot. An initial read from storage (SLOAD) costs 2100 gas.

It may be even cheaper to put in a struct so they can be read from storage once. But the second storage read (SLOAD) is only 100 gas so just using a struct may not save much more.

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,103 @@
...
+    uint256 public periodEnd;
+    uint256 public totalFunded;
```
naddison36: `totalFunded` and `totalRegistered` can easily fit in uint128 so both will only use one storage slot.

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,103 @@
...
+            periodStart = block.timestamp;
+            periodEnd = periodStart + PERIOD_DURATION;
+        }
+        totalFunded += amount;
+
+        emit Funded(msg.sender, amount);
+    }
+
+    /**
+     * @notice Allos user to register and transfer a given amount of MTA
```
naddison36: typo `Allows`

contracts/shared/MetaTokenRedeemer.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    /// @notice Redeems all user MTA balance for WETH at a fixed rate.
+    /// @return redeemedAmount The amount of WETH to receive.
+    function redeem() external returns (uint256 redeemedAmount) {
+        require(periodEnd <= block.timestamp, "Redeem period not started");
+        uint256 registeredAmount = balances[msg.sender];
+        require(registeredAmount > 0, "No balance");
+
+        // MTA and WETH both have 18 decimal points, no need for scaling.
+        redeemedAmount = (registeredAmount * totalRegistered) / totalFunded;
```
naddison36: this doesn't look right. Isn't it
`WETH redeemed = MTA registered * total WETH funded / total MTA registered`

naddison36 (changes requested):

contracts/shared/MetaTokenRedeemer.sol
doncesarts: Having it as a constant will block us from having unit tests on it.

naddison36 (approved):
Looks good

